PR #22, opened 2014-10-13
Added optional "remote ports forwardings" settings in advanced slave settings

PatriceG:
Hi,
I've added an optional "Remote Ports Forwadings" field in the advanced SSH field, that allows remote ports on the slave to be forwarded to the master.
I need this so that my SSH slaves do not need to know the IP address of Sonar and Sonar database.

If a comma-separated list of ports to forward is entered in the field then these ports are forwarded.
This is equivalent of -R localhost:port:localhost:port SSH options.

I hope you'll consider adding my changes to the official plugin code base.


stephenc:
If we are going to have this type of feature, then the port forwarding should have more control. Ideally I would see something like a `List<PortForwarding>` where you could specify the binding IPs on both sides as well as the ports... with perhaps environment variable expansion.

I can see utility in port forwarding, but the current implementation is just a hack implementation, and if the feature has merit then implementing it correctly has more merit.

oleg-nenashev:
@PatriceG Do you plan to finalize it?

PatriceG:
No I don't plan to finalize it sorry.
I don't need this anymore since we've moved our OSX build farm to TFS.

PR #25, opened 2015-07-16
Enable FindBugs XML reports in the plugin

oleg-nenashev:
- [x] Reconfigure pom.xml
- [x] Enable FindBugs on https://jenkins.ci.cloudbees.com/job/plugins/job/ssh-slaves-plugin/
- [x] Fix important issues

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,6 +14,11 @@
   <name>Jenkins SSH Slaves plugin</name>
   <url>http://wiki.jenkins-ci.org/display/JENKINS/SSH+Slaves+plugin</url>
 
+  <properties>
+    <findbugs-maven-plugin.version>3.0.1</findbugs-maven-plugin.version>
+    <findbugs.failOnError>true</findbugs.failOnError>
+  </properties>
+
   <developers>
     <developer>
       <id>stephenconnolly</id>
@@ -76,6 +81,12 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>annotations</artifactId>
+      <version>3.0.0</version>
+      <scope>provided</scope>
+    </dependency>
   </dependencies>
 
   <build>
@@ -100,6 +111,24 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <version>2.8</version>
+      </plugin>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>findbugs-maven-plugin</artifactId>
+        <version>${findbugs-maven-plugin.version}</version>
+        <configuration>
+          <xmlOutput>true</xmlOutput>
+          <failOnError>${findbugs.failOnError}</failOnError>
+        </configuration>
+        <executions>
+          <execution>
+            <id>run-findbugs</id>
+            <phase>verify</phase> 
+            <goals>
+              <goal>check</goal> 
+            </goals>
+          </execution>
+        </executions>
       </plugin>
     </plugins>
   </build>
--- a/src/main/java/hudson/plugins/sshslaves/SSHConnector.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHConnector.java
@@ -147,7 +147,9 @@ public class SSHConnector extends ComputerConnector {
         try {
             // only ever want from the system
             // lookup every time so that we always have the latest
-            StandardUsernameCredentials credentials = SSHLauncher.lookupSystemCredentials(credentialsId);
+            StandardUsernameCredentials credentials = 
+                    credentialsId == null ? null :
+                    SSHLauncher.lookupSystemCredentials(credentialsId);
             if (credentials != null) {
                 this.credentials = credentials;
                 return credentials;
@@ -259,8 +261,8 @@ public class SSHConnector extends ComputerConnector {
         this.prefixStartSlaveCmd = fixEmpty(prefixStartSlaveCmd);
         this.suffixStartSlaveCmd = fixEmpty(suffixStartSlaveCmd);
         this.launchTimeoutSeconds = launchTimeoutSeconds == null || launchTimeoutSeconds <= 0 ? null : launchTimeoutSeconds;
-        this.maxNumRetries = maxNumRetries == null || maxNumRetries <= 0 ? 0 : maxNumRetries;
-        this.retryWaitTime = retryWaitTime == null || retryWaitTime <= 0 ? 0 : retryWaitTime;
+        this.maxNumRetries = maxNumRetries != null && maxNumRetries > 0 ? maxNumRetries : 0;
+        this.retryWaitTime = retryWaitTime != null && retryWaitTime > 0 ? retryWaitTime : 0;
     }
 
     @Override
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -122,9 +122,12 @@ import java.util.regex.Pattern;
 
 import static com.cloudbees.plugins.credentials.CredentialsMatchers.*;
 import com.cloudbees.plugins.credentials.common.StandardUsernameListBoxModel;
+import edu.umd.cs.findbugs.annotations.CheckForNull;
 import static hudson.Util.*;
 import hudson.model.Computer;
 import hudson.security.AccessControlled;
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.Charset;
 import static java.util.logging.Level.*;
 
 /**
@@ -457,8 +460,8 @@ public class SSHLauncher extends ComputerLauncher {
         this.prefixStartSlaveCmd = fixEmpty(prefixStartSlaveCmd);
         this.suffixStartSlaveCmd = fixEmpty(suffixStartSlaveCmd);
         this.launchTimeoutSeconds = launchTimeoutSeconds == null || launchTimeoutSeconds <= 0 ? null : launchTimeoutSeconds;
-        this.maxNumRetries = maxNumRetries == null || maxNumRetries <= 0 ? 0 : maxNumRetries;
-        this.retryWaitTime = retryWaitTime == null || retryWaitTime <= 0 ? 0 : retryWaitTime;
+        this.maxNumRetries = maxNumRetries != null && maxNumRetries > 0 ? maxNumRetries : 0;
+        this.retryWaitTime = retryWaitTime != null && retryWaitTime > 0 ? retryWaitTime : 0;
     }
 
     /** @deprecated Use {@link #SSHLauncher(String, int, StandardUsernameCredentials, String, String, JDKInstaller, String, String)} instead. */
@@ -483,7 +486,8 @@ public class SSHLauncher extends ComputerLauncher {
         try {
             // only ever want from the system
             // lookup every time so that we always have the latest
-            StandardUsernameCredentials credentials = SSHLauncher.lookupSystemCredentials(credentialsId);
+            StandardUsernameCredentials credentials = credentialsId != null ? 
+                    SSHLauncher.lookupSystemCredentials(credentialsId) : null;
             if (credentials != null) {
                 this.credentials = credentials;
                 return credentials;
@@ -673,11 +677,16 @@ public class SSHLauncher extends ComputerLauncher {
      *
      * @return the remote root workspace (without trailing slash).
      */
+    @CheckForNull
     private static String getWorkingDirectory(SlaveComputer computer) {
         return getWorkingDirectory(computer.getNode());
     }
 
-    private static String getWorkingDirectory(Slave slave) {
+    @CheckForNull
+    private static String getWorkingDirectory(@CheckForNull Slave slave) {
+        if (slave == null) {
+            return null;
+        }
         String workingDirectory = slave.getRemoteFS();
         while (workingDirectory.endsWith("/")) {
             workingDirectory = workingDirectory.substring(0, workingDirectory.length() - 1);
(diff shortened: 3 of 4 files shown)


oleg-nenashev:
@reviewbybees
Cleaned up all issues discovered by FindBugs

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -787,8 +806,7 @@ protected String resolveJava(SlaveComputer computer, TaskListener listener) thro
                 try {
                     tried.add(javaCommand);
                     String java = checkJavaVersion(listener, javaCommand);
```
KostyaSha: just return it?

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -852,7 +871,14 @@ private EnvVars getEnvVars(DescribableList<NodeProperty<?>, NodePropertyDescript
     private void verifyNoHeaderJunk(TaskListener listener) throws IOException, InterruptedException {
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
         connection.exec("true",baos);
-        String s = baos.toString();
+        final String s;
+        //TODO: Seems we need to retrieve the encoding from the connection destination
+        try {
+            s = baos.toString(Charset.defaultCharset().name());
+        } catch (UnsupportedEncodingException ex) { // Should not happen
+            throw new IOException("Default encoding is unsupported");
```
KostyaSha: `, ex`?

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -883,8 +909,14 @@ private String attemptToInstallJDK(TaskListener listener, String workingDirector
...
         //        (this one is from MKS)
-
-        String uname = unameOutput.toString();
+        
+        //TODO: Seems we need to retrieve the encoding from the connection destination
+        final String uname;
+        try {
+            uname = unameOutput.toString(Charset.defaultCharset().name());
+        } catch (UnsupportedEncodingException ex) { // Should not happen
+            throw new IOException("Default encoding is unsupported");
```
KostyaSha: `, ex`?

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -1071,13 +1103,16 @@ protected void reportEnvironment(TaskListener listener) throws IOException, Inte
...
+    @NonNull
     private String checkJavaVersion(TaskListener listener, String javaCommand) throws IOException, InterruptedException {
         listener.getLogger().println(Messages.SSHLauncher_CheckingDefaultJava(getTimestamp(),javaCommand));
         StringWriter output = new StringWriter();   // record output from Java
 
         ByteArrayOutputStream out = new ByteArrayOutputStream();
         connection.exec(javaCommand + " "+getJvmOptions() + " -version",out);
-        BufferedReader r = new BufferedReader(new InputStreamReader(new ByteArrayInputStream(out.toByteArray())));
+        //TODO: Seems we need to retrieve the encoding from the connection destination
+        BufferedReader r = new BufferedReader(new InputStreamReader(
```
KostyaSha: wouldn't it cause resource leak?

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -457,8 +460,9 @@ public SSHLauncher(String host, int port, StandardUsernameCredentials credential
         this.prefixStartSlaveCmd = fixEmpty(prefixStartSlaveCmd);
         this.suffixStartSlaveCmd = fixEmpty(suffixStartSlaveCmd);
         this.launchTimeoutSeconds = launchTimeoutSeconds == null || launchTimeoutSeconds <= 0 ? null : launchTimeoutSeconds;
-        this.maxNumRetries = maxNumRetries == null || maxNumRetries <= 0 ? 0 : maxNumRetries;
-        this.retryWaitTime = retryWaitTime == null || retryWaitTime <= 0 ? 0 : retryWaitTime;
+        final Integer zero = 0;
```
KostyaSha: `Integer.valueOf(0)`?

stephenc: Any constants in the range -128..127 will be subst by the compiler anyway

src/main/java/hudson/plugins/sshslaves/SSHConnector.java
```diff
@@ -259,8 +261,9 @@ public SSHConnector(int port, StandardUsernameCredentials credentials, String us
         this.prefixStartSlaveCmd = fixEmpty(prefixStartSlaveCmd);
         this.suffixStartSlaveCmd = fixEmpty(suffixStartSlaveCmd);
         this.launchTimeoutSeconds = launchTimeoutSeconds == null || launchTimeoutSeconds <= 0 ? null : launchTimeoutSeconds;
-        this.maxNumRetries = maxNumRetries == null || maxNumRetries <= 0 ? 0 : maxNumRetries;
-        this.retryWaitTime = retryWaitTime == null || retryWaitTime <= 0 ? 0 : retryWaitTime;
+        final Integer zero = 0;
+        this.maxNumRetries = maxNumRetries == null || maxNumRetries <= 0 ? zero : maxNumRetries;
+        this.retryWaitTime = retryWaitTime == null || retryWaitTime <= 0 ? zero : retryWaitTime;
```
stephenc: I would have gone with `retryWaitTime == null ? 0 : Math.max(0, retryWaitTime)` or `Math.max(0, retryWaitTime == null? 0 : retryWaitTime)` or `retryWaitTime == null || retryWaitTime <= 0 ? 0 : (int)retryWaitTime` adding a local constant seems insane just to stop complaints about covariant returns from an elvis operator

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
oleg-nenashev: There is no resource leak here, because we don't use any file resources here

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
KostyaSha: ah, right. java compiler is also smart as gcc

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
stephenc: not with a byte array stream as there are no external resources to leak... but it could cause complaints of a resource leak...

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -1216,7 +1252,7 @@ public synchronized void afterDisconnect(SlaveComputer slaveComputer, TaskListen
                     if (sftpClient == null) {// system without SFTP
                         try {
                             connection.exec("rm " + fileName, listener.getLogger());
-                        } catch (Exception x) {
+                        } catch (Throwable x) {
```
stephenc: What about letting `Error` fall through. Seems like a change above and beyond findbugs

pom.xml
```diff
@@ -14,6 +14,11 @@
   <name>Jenkins SSH Slaves plugin</name>
   <url>http://wiki.jenkins-ci.org/display/JENKINS/SSH+Slaves+plugin</url>
 
+  <properties>
+    <findbugs-maven-plugin.version>3.0.1</findbugs-maven-plugin.version>
```
stephenc: I *hate* people using properties for a dependency/plugin version that is used once and only once in the pom. Using properties for versions is really not a good plan unless you expressly know exactly what you are doing and why you are doing it.

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
oleg-nenashev: @stephenc 
It does not cause complains at least in FindBugs 3.0.1

pom.xml
```diff
@@ -14,6 +14,11 @@
...
+    <findbugs-maven-plugin.failOnError>true</findbugs-maven-plugin.failOnError>
```
stephenc: The [user property](http://gleclaire.github.io/findbugs-maven-plugin/check-mojo.html#failOnError) is `findbugs.failOnError` do not hijack the expected property and move it to a different name

pom.xml
```diff
@@ -101,6 +112,25 @@
...
         <version>2.8</version>
       </plugin>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>findbugs-maven-plugin</artifactId>
+        <version>${findbugs-maven-plugin.version}</version>
+        <configuration>
+          <xmlOutput>true</xmlOutput>
+          <findbugsXmlWithMessages>true</findbugsXmlWithMessages>
+          <failOnError>${findbugs-maven-plugin.failOnError}</failOnError>
```
stephenc: Do not hijack the existing user property. If a parent has forced this to `true` or `false` and you want to switch it back, then use `${findbugs.failOnError}` not a random property name that does not fit in with the existing maven plugin documentation

pom.xml
KostyaSha: it exists in upstream pom

src/main/java/hudson/plugins/sshslaves/SSHConnector.java
stephenc: Actually, wouldn't the correct thing be to change the field type to `int` as that is the default value when `null` and that would remove the boxing/unboxing

pom.xml
stephenc: Ugh! Yes I forgot that the upstream pom does every kind of wrong under the sun

pom.xml
KostyaSha: i think @oleg-nenashev followed existing style with controlling jenkins build processes with properties that come from 'pom' project. Handling env properties from profile also makes sense but seems doesn't fit idea with enforcing it.

pom.xml
KostyaSha: Well, it not so bad imho in comparison to redundant ancient pom project https://github.com/jenkinsci/pom

pom.xml
KostyaSha: @oleg-nenashev maybe just remove this variable and set true in configuration section?

KostyaSha: @stephenc please check maillist i sent my thought about this variables. Maybe you can propose something better.

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
oleg-nenashev: We got into several critical failures in buggy Jenkins extensions due to catching Exceptions and Missing throwables. Probably for this code it makes sense to accept Errors

src/main/java/hudson/plugins/sshslaves/SSHConnector.java
oleg-nenashev: The field is public => no easy way back to <code>int</code>

oleg-nenashev: @stephenc 
I'll change the order of conditions to <code>this.maxNumRetries = maxNumRetries != null && maxNumRetries > 0 ? maxNumRetries : 0;</code>. It will prevent a pointless boxing-unboxing cycle, because Integer will be taken as a return type of the operator

oleg-nenashev:
@stephenc @KostyaSha 
I've fixed your comments

pom.xml
stephenc: There is nothing to think about IMHO. the existing property used to control this configuration option is called `findbugs.failOnError` as per the plugins own [documentation](http://gleclaire.github.io/findbugs-maven-plugin/check-mojo.html#failOnError). 

I might have personal complaints about that specific plugin having picked a crummy name for the property... but that does not change the fact that the convention of all maven users is to use the property described in the plugin's documentation.

So the name should be `findbugs.failOnError` (even though I do not like that specific property name on so many levels... those complaints are to be addressed with the findbugs plugin maintainers)

stephenc:
:bug: due to not following maven conventions for `findbugs.failOnError` property name

pom.xml
KostyaSha: +1 for renaming or just hardcode it in configuration section.

oleg-nenashev: Agreed

stephenc:
:bee:

KostyaSha:
:+1:

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -1012,10 +1042,10 @@ private void copySlaveJar(TaskListener listener, String workingDirectory) throws
                         os.close();
                     }
                     listener.getLogger().println(Messages.SSHLauncher_CopiedXXXBytes(getTimestamp(), slaveJar.length));
-                } catch (Exception e) {
+                } catch (Throwable e) {
```
stephenc: You should let `Error` passthrough as is. Escalating to `Throwable` without a passthrough for `Error` is wrong

stephenc: Also changing from Exception to Throwable should be a separate PR IMHO

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -1216,8 +1250,11 @@ public synchronized void afterDisconnect(SlaveComputer slaveComputer, TaskListen
                     if (sftpClient == null) {// system without SFTP
                         try {
                             connection.exec("rm " + fileName, listener.getLogger());
-                        } catch (Exception x) {
+                        } catch (Throwable x) {
                             x.printStackTrace(listener.error(Messages.SSHLauncher_ErrorDeletingFile(getTimestamp())));
+                            if (x instanceof Error) {
```
stephenc: Ugh! nope... that's not the way to do it

```
} catch (Error x) { 
  throw x;
} catch (Throwable x) {
```
is the way to do it

stephenc:
:bug: exception mis-handling of `Error`s

oleg-nenashev:
@stephenc 
Fixed the exceptions

stephenc:
:bee:

oleg-nenashev:
@reviewbybees done
Squashing commits

PR #27, opened 2015-11-18
[JENKINS-29412] Reject Java < 1.7

daniel-beck:
Otherwise Jenkins will happily try to run its `slave.jar` on outdated Java versions (including plugin-provided Java 6 from earlier plugin versions).


oleg-nenashev:
:-1: , because the core dependency is 1.580.1 This core supports Java6

daniel-beck:
So? We're not refusing to run, we just don't choose a JRE that may not actually work. This has been changed to download a more recent version a long time ago.

Of course this could raise the dependency, but wouldn't that be even worse for users < 1.612/1.625.x?

daniel-beck:
@jglick Yes.

daniel-beck:
@oleg-nenashev Are you still opposed to this approach?

daniel-beck:
@oleg-nenashev Ping

oleg-nenashev:
The plugin effectively requires 1.625 due to the transient dependencies. Just bump Jenkins version in the properties, and it will be perfectly fine.

Though we will still need to do something with Java 8 strict requirement after that

daniel-beck:
@oleg-nenashev 

> Though we will still need to do something with Java 8 strict requirement after that

Could you please respond to https://github.com/jenkinsci/ssh-slaves-plugin/pull/27#issuecomment-204187486?

daniel-beck:
I asked Oleg directly and he maintains his rejection of this PR, so pointless to adapt this to Java 8.

As nobody else cared enough to merge this in >1.5 years, and it now has conflicts, I'm closing this.

oleg-nenashev:
We have discussed the approach with @daniel-beck offline. The current plan:

- @oleg-nenashev to come up with a PR, which dynamically detects Jenkins core versions and makes decisions about Java requirements
- @oleg-nenashev to scanvenge tests from this pull request

jglick:
Why not just use `ComputerLauncher.checkJavaVersion`, as the TODO comment here suggests?

PR #28, opened 2016-01-15
Meaningful thread name for callable establishing SSH connection

ydubreuil:
This is needed to diagnose slave connection initialization issues. Currently, SSH connection initialization is done in a  `Callable` running inside an anonymous and dedicated `ExecutorService`. If something goes wrong and connection timeout is set to 0, connection thread can get stuck and never killed. Without this, we can't find which slave is involved.

Bumping core dependency was required to use `NamingThreadFactory`.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.532</version>
+    <version>1.554</version>
   </parent>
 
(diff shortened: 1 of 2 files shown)


amuniz:
I'm :+1: with the change, but please file a JIRA issue to link to.

andresrc:
:+1: , will bee when linked to JIRA.

oleg-nenashev:
@ydubreuil It is not mergeable anymore

PR #29, opened 2016-01-18
[FIXED JENKINS-32489] "Prefix/Suffix Start Slave Command" does not add a space before/after the cmd

fbelzunc:
Add the suffix below into a ssh-slave.

``` 
2> >(tee -a slave-stderr.log 1>&2)
```
Then, the following happens:

```
[01/18/16 08:32:37] [SSH] Copied 478,051 bytes.
Expanded the channel window size to 4MB
[01/18/16 08:32:37] [SSH] Starting slave process: cd "/home/vagrant" && java  -jar slave.jar2> >(tee -a slave-stderr.log 1>&2)
Unable to access jarfile slave.jar2
hudson.util.IOException2: Slave JVM has terminated. Exit code=1
	at hudson.plugins.sshslaves.SSHLauncher.startSlave(SSHLauncher.java:953)
	at hudson.plugins.sshslaves.SSHLauncher.access$400(SSHLauncher.java:133)
	at hudson.plugins.sshslaves.SSHLauncher$2.call(SSHLauncher.java:711)
	at hudson.plugins.sshslaves.SSHLauncher$2.call(SSHLauncher.java:696)
	at java.util.concurrent.FutureTask.run(FutureTask.java:262)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.io.EOFException: unexpected stream termination
	at hudson.remoting.ChannelBuilder.negotiate(ChannelBuilder.java:331)
	at hudson.remoting.ChannelBuilder.build(ChannelBuilder.java:280)
	at hudson.slaves.SlaveComputer.setChannel(SlaveComputer.java:371)
	at hudson.plugins.sshslaves.SSHLauncher.startSlave(SSHLauncher.java:945)
	... 7 more
[01/18/16 08:32:37] Launch failed - cleaning up connection
```
The same happens with the prefix.

@reviewbybees

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -965,7 +965,7 @@ public class SSHLauncher extends ComputerLauncher {
         String cmd = "cd \"" + workingDirectory + "\" && " + java + " " + getJvmOptions() + " -jar slave.jar";
 
         //This will wrap the cmd with prefix commands and suffix commands if they are set.
-        cmd = getPrefixStartSlaveCmd() + cmd + getSuffixStartSlaveCmd();
+        cmd = getPrefixStartSlaveCmd() + " " + cmd + " " + getSuffixStartSlaveCmd();
 
         listener.getLogger().println(Messages.SSHLauncher_StartingSlaveProcess(getTimestamp(), cmd));
         session.execCommand(cmd);


oleg-nenashev:
It breaks use-cases when prefix and suffix somehow contribute to the main command. IIRC we used it in order to arbitrary select slave.jar versions. E.g. suffix=".2.34" was leading to the "java -jar slave.jar.2.34"

So this is a breaking change. :bug: until this change is justified. Why don't you just add a space to the suffix?

fbelzunc:
> IIRC we used it in order to arbitrary select slave.jar versions.

I didn't know this, although IIRC now we don't have now such a nomenclature in `slave.jar`

> Why don't you just add a space to the suffix?

Yes, you can do this. But I saw that to need to add a space on the GUI before such kind of commands is very confusing for Jenkins users. When you hit this problem, from the user point of view it sounds more a bug than really something it is intended.

oleg-nenashev:
> I didn't know this, although IIRC now we don't have now such a nomenclature in slave.jar

Actually we have. At my previous instance we had to maintain our own custom fork of remoting, hence multi-version support was critical to our test installations. The use-case can be implemented in a different way, but that's what we had on early testing stages.

fbelzunc:
@oleg-nenashev Should I close this pull request then and JENKINS-32489 accordingly?

oleg-nenashev:
@fbelzunc Maybe makes sense to wait for a feedback from others. If they find this change useful, we will be able to consider alternate fixes allowing to retain the original behavior somehow

oleg-nenashev:
@fbelzunc Seems there's no feedback. I propose to close it

fbelzunc:
Yes, I agree. I think current implementation is a bit confused for non-advanced used, but I understand this might break instances.

PR #31, opened 2016-03-22
[JENKINS-33720] Upgrade to plugin pom 2.4 and bump baseline

armfergom:
Included in this PR:

- [x] Upgrade to plugin parent pom 2.4
- [x] Bump baseline to 1.580.1. With 1.532, project failed to build (with Java8) due to a bug in XStream [that was fixed in XStream 1.4.6](http://x-stream.github.io/jira/746/). 1.532 includes version 1.4.4 of XStream while 1.580.1 includes 1.4.7.
- [x] Fix FindBugs errors after upgrade
- [x] Fix Javadoc errors after upgrade
- [x] Fix injected tests errors after upgrade

PCT runs successfully after the upgrade. 

@reviewbybees esp. @stephenc

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.532</version>
+    <version>2.4</version>
   </parent>
 
   <artifactId>ssh-slaves</artifactId>
@@ -17,6 +17,8 @@
   <properties>
     <findbugs-maven-plugin.version>3.0.1</findbugs-maven-plugin.version>
     <findbugs.failOnError>true</findbugs.failOnError>
+    <jenkins.version>1.580.1</jenkins.version>
+    <java.level>6</java.level>
   </properties>
 
   <developers>
@@ -67,26 +69,6 @@
       <artifactId>ssh-credentials</artifactId>
       <version>1.6.1</version>
     </dependency>
-    <!-- jenkins dependencies -->
-    <!-- test dependencies -->
-    <dependency>
-      <groupId>org.jenkins-ci.main</groupId>
-      <artifactId>jenkins-test-harness</artifactId>
-      <version>${project.parent.version}</version>
-      <scope>test</scope>
-      <exclusions>
-        <exclusion>
-          <groupId>${project.groupId}</groupId>
-          <artifactId>${project.version}</artifactId>
-        </exclusion>
-      </exclusions>
-    </dependency>
-    <dependency>
-      <groupId>com.google.code.findbugs</groupId>
-      <artifactId>annotations</artifactId>
-      <version>3.0.0</version>
-      <scope>provided</scope>
-    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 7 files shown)


andresrc:
:bee:

abayer:
:bee:

pom.xml
```diff
@@ -94,43 +76,13 @@
       <plugin>
         <groupId>org.jenkins-ci.tools</groupId>
         <artifactId>maven-hpi-plugin</artifactId>
-        <version>1.106</version>
         <configuration>
           <systemProperties>
             <hudson.bundled.plugins>${basedir}/work/plugins/ssh-slaves.hpl</hudson.bundled.plugins>
```
jglick: [This hack](https://github.com/jenkinsci/jenkins/blob/e2dccc35572b64f5e602a6b1b3efb517c578872c/core/src/main/java/hudson/LocalPluginManager.java#L63) is probably obsolete. Try deleting the `maven-hpi-plugin` configuration, then doing a

    git clean -fdx && mvn hpi:run

and verify that http://localhost:8080/jenkins/pluginManaged/installed shows the this plugin as `1.11-SNAPSHOT` as expected.

src/main/java/hudson/plugins/sshslaves/SSHConnector.java
```diff
@@ -279,7 +279,8 @@ public String getDisplayName() {
         }
 
         public ListBoxModel doFillCredentialsIdItems(@AncestorInPath ItemGroup context) {
-            if (!(context instanceof AccessControlled ? (AccessControlled) context : Jenkins.getInstance()).hasPermission(Computer.CONFIGURE)) {
+            AccessControlled _context = (context instanceof AccessControlled ? (AccessControlled) context : Jenkins.getInstance());
+            if (_context != null && !_context.hasPermission(Computer.CONFIGURE)) {
```
jglick: :bug: this would open a security hole: a direct call to `/descriptorById/hudson.plugins.sshslaves.SSHConnector/fillCredentialsIdItems` from a user with only `Jenkins.READ` permission (perhaps anonymous) could enumerate credentials.

I think you meant

```java
if (_context == null || !_context.hasPermission(Computer.CONFIGURE)) {
```

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -1472,7 +1472,8 @@ public String getHelpFile(String fieldName) {
         public ListBoxModel doFillCredentialsIdItems(@AncestorInPath ItemGroup context,
                                                      @QueryParameter String host,
                                                      @QueryParameter String port) {
-            if (!(context instanceof AccessControlled ? (AccessControlled) context : Jenkins.getInstance()).hasPermission(Computer.CONFIGURE)) {
+            AccessControlled _context = (context instanceof AccessControlled ? (AccessControlled) context : Jenkins.getInstance());
+            if (_context != null && !_context.hasPermission(Computer.CONFIGURE)) {
```
jglick: :bug: ditto

pom.xml
armfergom: Checked and works. Removed this plugin from the plugin pom. Thanks.

src/main/java/hudson/plugins/sshslaves/SSHConnector.java
armfergom: I am not sure how this would _add_ a security hole, but it's true that if `Jenkins.getInstance()` returns null, the behaviour should be to just return the empty `ListBoxModel`, so I have changed it anyway.

jglick:
> I am not sure how this would _add_ a security hole

My original fear was that an attacker could

    curl http://jenkins/descriptorById/hudson.plugins.sshslaves.SSHConnector/fillCredentialsIdItems

and that you would then skip the security check. On closer inspection, this would actually still pass a `Jenkins` for the `context` (as this is an `ItemGroup`), and even if it did not, the `instanceof` check would be false and we would fall back to `Jenkins.getInstance()`. There is still an edge case: if an attacker makes this call _as Jenkins is shutting down_, it may be possible for the HTTP request to start to be handled right before the shutdown, then this web method is called, at which point Stapler argument binding might not pick up the ancestor, and `Jenkins.getInstance() == null`, so `_context == null` too, bypassing the check and granting access to secrets (whereas the original code would merely throw a `NullPointerException` and leak nothing).

Unlikely to be exploitable, but it is best to err on the side of caution when performing security checks: if something you need is missing, fail. You can also feel free to call `Jenkins.getActiveInstance()` from a web method: either Jenkins is up and running, or we should not be handling a request now and we might as well fail early.

jglick:
:bee:

src/main/java/hudson/plugins/sshslaves/SSHConnector.java
stephenc: `Jenkins.getInstance()` can never be null in a web UI invoked method such as this

andresrc:
:bee:

armfergom:
@reviewbybees done

PR #40, opened 2016-12-19
[JENKINS-30561] Attempt to install JDK if min java requirements are not met

e-g-hategan:
https://issues.jenkins-ci.org/browse/JENKINS-30561

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -144,7 +144,9 @@ public class SSHLauncher extends ComputerLauncher {
     public static final SchemeRequirement SSH_SCHEME = new SchemeRequirement("ssh");
 
 
-    public static final String JDKVERSION = "jdk-7u80";
+    public static final int JDK_MAJOR_VERSION = 7;
+    public static final int JDK_MINOR_VERSION = 80;
+    public static final String JDKVERSION = "jdk-" + JDK_MAJOR_VERSION + "u" + JDK_MINOR_VERSION;
     public static final String DEFAULT_JDK = JDKVERSION + "-oth-JPR";
 
     /**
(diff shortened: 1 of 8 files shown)


e-g-hategan:
Just clarifying: do you mean I should bump jenkins.version to latest 2.50, java.level to 7 and address any upgrade related issues?

oleg-nenashev:
Definitely not to `2.50`. I am not sure what should be used exactly. `credentials` plugin dependency in the plugin already requires 1.625.3, hence I doubt this plugin actually can support Java6 anyway (CC @daniel-beck, @olivergondza , @stephenc since they are involved into the related changes).

So I suppose `1.625.3` is the best possible target

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -1176,6 +1178,10 @@ protected String checkJavaVersion(final PrintStream logger, String javaCommand,
         return null;
     }
 
+    private double requiredJavaVersion() {
+        return Double.valueOf("1." + JDK_MAJOR_VERSION);
```
olivergondza: How about base this on a minimal version that is supported by jenkins version running on master (provided we can get that requirement at runtime)? Or at least use the version running on master as the minimum.

e-g-hategan: That's a good idea.

oleg-nenashev: It can be hardcoded by checking the remoting version

e-g-hategan: Based on Jenkins.getInstance().getVersion() we have 3 possible outcomes: 
before 1.610 then java 6
before 2.5x: java 7
afterwards java 8 (will happen soon; confirmed on the IRC channel).

oleg-nenashev: Something like that. But I am not sure this plugin can wok on Java 6 due to the credentials dependency

monkz: as jenkins.io offers a jdk list - i recommend using this list to choose a JDK version

PR #43, opened 2017-02-24
[JENKINS-43481] Bump JDKVERSION to jdk8-u121

monkz:
As jdk7 reached end of life, i would recommend to bump the version to a rather recent one.

We might discuss further, if we expose JDKVERSION via "Manage Jenkins".
Like "latest", "please type jdkversion here/choose from these jdks" and a "stable version".
The latter one would behave like a final JDKVERSION.

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -144,7 +144,7 @@ public class SSHLauncher extends ComputerLauncher {
     public static final SchemeRequirement SSH_SCHEME = new SchemeRequirement("ssh");
 
 
-    public static final String JDKVERSION = "jdk-7u80";
+    public static final String JDKVERSION = "jdk-8u181";
     public static final String DEFAULT_JDK = JDKVERSION + "-oth-JPR";
 
     // Some of the messages observed in the wild:
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -144,7 +144,7 @@ public class SSHLauncher extends ComputerLauncher {
     public static final SchemeRequirement SSH_SCHEME = new SchemeRequirement("ssh");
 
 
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
It would automagically solve https://issues.jenkins-ci.org/browse/JENKINS-43481 though there is some risk for older Jenkins cores. But I think it is a reasonable step forward

jglick:
> there is some risk for older Jenkins cores

What risk? We certainly recommend running the agent on the newest official JRE even if Jenkins does not strictly require it.

jglick:
cf. #27

jglick:
#40 is more general, but this seems safer.

jglick:
Fix verified by:

* `hpi:run`
* `docker run --rm --name sshd -p 127.0.0.1:2222:22 rastasheep/ubuntu-sshd`
* create username/password credentials `root:root`
* configure SSH agent with
```xml
<?xml version="1.0" encoding="UTF-8"?>
<slave>
  <name>ssh</name>
  <description></description>
  <remoteFS>/tmp/agent</remoteFS>
  <numExecutors>1</numExecutors>
  <mode>NORMAL</mode>
  <retentionStrategy class="hudson.slaves.RetentionStrategy$Always"/>
  <launcher class="hudson.plugins.sshslaves.SSHLauncher" plugin="ssh-slaves@1.17-SNAPSHOT">
    <host>localhost</host>
    <port>2222</port>
    <credentialsId>…</credentialsId>
    <maxNumRetries>0</maxNumRetries>
    <retryWaitTime>0</retryWaitTime>
    <sshHostKeyVerificationStrategy class="hudson.plugins.sshslaves.verifiers.ManuallyTrustedKeyVerificationStrategy">
      <requireInitialManualTrust>false</requireInitialManualTrust>
    </sshHostKeyVerificationStrategy>
  </launcher>
  <label></label>
  <nodeProperties/>
  <userId>anonymous</userId>
</slave>
```
* launch, check that
   * log says JDK was downloaded & installed
   * log says `slave.jar` was copied
   *  `/computer/*/systemInfo` says `java.version` is `1.8.0_121`

PR #44, opened 2017-03-08
Improve exception reporting

olivergondza:


src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -801,15 +802,15 @@ private void cleanupConnection(TaskListener listener) {
      * return javaPath if specified in the configuration.
      * Finds local Java, and if none exist, install one.
      */
-    protected String resolveJava(SlaveComputer computer, TaskListener listener) throws InterruptedException, IOException2 {
+    protected String resolveJava(SlaveComputer computer, TaskListener listener) throws InterruptedException, IOException {
```
olivergondza: Never used in public plugins.

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -989,14 +990,14 @@ private void startSlave(SlaveComputer computer, final TaskListener listener, Str
...
         } catch (InterruptedException e) {
             session.close();
-            throw new IOException2(Messages.SSHLauncher_AbortedDuringConnectionOpen(), e);
+            throw new IOException(Messages.SSHLauncher_AbortedDuringConnectionOpen(), e);
         } catch (IOException e) {
             try {
                 // often times error this early means the JVM has died, so let's see if we can capture all stderr
                 // and exit code
-                throw new IOException2(getSessionOutcomeMessage(session,false),e);
+                throw new AbortException(getSessionOutcomeMessage(session,false));
```
olivergondza: There is no reason to obscure the the friendly cause message with irrelevant stacktrace.

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
oleg-nenashev: Also I see no usages in private-source plugins, to which I have access.
But maybe you want to create a new method and use the original one as a wrapper. NIT anyway

olivergondza: Does not seem to be worth the hassle. Only clients impacted are those that handle `IOException2` and `IOException` differently - which does not really makes sense. Fortunately, this will break source compatibility pointing them to the problem.

PR #47, opened 2017-03-17
Use apache sshd instead of insecure trilead-ssh2

paladox:


paladox:
Oh, sorry i did this because i thought it was easy and that trilead-ssh2 wasent being updated. Now that it is i can abandon this :)

PR #51, opened 2017-05-06
[JENKINS-42959] Support Ed25519 and EDDSA host keys

reedloden:
* Trilead just gained support for Ed25519 and EDDSA algorithms, so update
   the hardcoded list in the manual key verification step to support them.
* Add a isClass() method to allow using classes in a newer Trilead version
   while still retaining backwards compatibility with old Jenkins versions.
* Fix a few typos that I noticed while adding this.

PR #52, opened 2017-05-07
Update ssh slaves to build against java 8 and jenkins 2.58

paladox:


paladox:
Could @mc1arke or @oleg-nenashev or @jglick please review?

paladox:
@oleg-nenashev it may be required but not really sure. I think i found a workaround here https://github.com/jenkinsci/ssh-slaves-plugin/pull/51 though that workaround is untested. Though i hope the workaround works :)

rbu:
I'm not on LTS for other features that I need, so this issue is pretty critical for me. I'd appreciate it if you could merge one of these PRs or push out a release with your proposed solution.

PR #53, opened 2017-05-11
Fix a few typos and add a translation entry for unknown algorithm

reedloden:
Split off from #51.


oleg-nenashev:
Merged with master

PR #54, opened 2017-05-16
[FIXED JENKINS-42959] Specify preferred host keys during connect

mc1arke:
The remaining fix for JENKINS-42959: adds support for additional host key algorithms where Jenkins contains a newer version of Trilead, and specify the preferred host key algorithm for connecting in-case Jenkins has an old key format saved as the known key.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,8 +16,8 @@
   <description>Allows to launch agents over SSH, using a Java implementation of the SSH protocol</description>
 
   <properties>
-    <jenkins.version>1.609.1</jenkins.version>
-    <java.level>6</java.level>
+    <jenkins.version>1.625</jenkins.version>
+    <java.level>7</java.level>
     <jenkins-test-harness.version>2.18</jenkins-test-harness.version>
   </properties>
 
@@ -62,6 +62,13 @@
 
   <dependencies>
     <!-- regular dependencies -->
+    <dependency>
+      <groupId>org.jenkins-ci</groupId>
+      <artifactId>trilead-ssh2</artifactId>
+      <version>build217-jenkins-9</version>
+      <scope>provided</scope>
+      <!-- we only need the newer version for testing, we use the bundled version during execution -->
+    </dependency>
     <!-- plugin dependencies -->
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -65,7 +65,7 @@
     <dependency>
       <groupId>org.jenkins-ci</groupId>
       <artifactId>trilead-ssh2</artifactId>
-      <version>build217-jenkins-9</version>
+      <version>build-217-jenkins-11</version>
       <scope>provided</scope>
       <!-- we only need the newer version for testing, we use the bundled version during execution -->
     </dependency>
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -786,6 +786,7 @@ public class SSHLauncher extends ComputerLauncher {
             public Boolean call() throws InterruptedException {
                 Boolean rval = Boolean.FALSE;
                 try {
+                    connection.setServerHostKeyAlgorithms(sshHostKeyVerificationStrategy.getPreferredKeyAlgorithms(computer));
 
                     openConnection(listener, computer);
 
--- a/src/main/java/hudson/plugins/sshslaves/verifiers/JenkinsTrilead9VersionSupport.java
+++ b/src/main/java/hudson/plugins/sshslaves/verifiers/JenkinsTrilead9VersionSupport.java
@@ -1,0 +1,39 @@
+package hudson.plugins.sshslaves.verifiers;
+
+import com.trilead.ssh2.signature.KeyAlgorithm;
+import com.trilead.ssh2.signature.KeyAlgorithmManager;
+import hudson.plugins.sshslaves.Messages;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * @author Michael Clarke
+ */
+class JenkinsTrilead9VersionSupport extends TrileadVersionSupportManager.TrileadVersionSupport {
+
+    @Override
+    public String[] getSupportedAlgorithms() {
+        List<String> algorithms = new ArrayList<>();
+        for (KeyAlgorithm<?, ?> algorithm : KeyAlgorithmManager.getSupportedAlgorithms()) {
+            algorithms.add(algorithm.getKeyFormat());
+        }
+        return algorithms.toArray(new String[algorithms.size()]);
+    }
+
+    @Override
+    public HostKey parseKey(String algorithm, byte[] keyValue) {
+        for (KeyAlgorithm<?, ?> keyAlgorithm : KeyAlgorithmManager.getSupportedAlgorithms()) {
+            try {
+                if (keyAlgorithm.getKeyFormat().equals(algorithm)) {
+                    keyAlgorithm.decodePublicKey(keyValue);
+                    return new HostKey(algorithm, keyValue);
+                }
+            } catch (IOException ex) {
+                throw new IllegalArgumentException(Messages.ManualKeyProvidedHostKeyVerifier_KeyValueDoesNotParse(algorithm), ex);
+            }
+        }
+        throw new IllegalArgumentException("Unexpected key algorithm " + algorithm);
+    }
+}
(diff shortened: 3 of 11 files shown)


pom.xml
```diff
@@ -62,6 +62,13 @@
 
   <dependencies>
     <!-- regular dependencies -->
+    <dependency>
+      <groupId>org.jenkins-ci</groupId>
+      <artifactId>trilead-ssh2</artifactId>
+      <version>build217-jenkins-9</version>
```
paladox: Should be build217-jenkins-10 please.

mc1arke: Nope. We want the oldest version of Trilead that provides the new API. This version is only used during the compile/test phase, not in the package/release phase so the actual version here doesn't impact what end users get, but it needs to be a version that provides the API required in the PR without introducing other API that may not exist in versions bundled in other Jenkins releases.

oleg-nenashev: Ideally should be replaced by API plugin when it's ready

paladox:
Tests failures seem to be because the build is set as unstable.

mc1arke:
The build is unstable due to the test failure, which is what is blocking this PR from being sent for full approval.

pom.xml
mc1arke: Agreed. I'll pick up the API plugin work as soon as I get to the bottom of any major outstanding Trilead issues.

mc1arke:
Tests fixed. Feel free to chip in with reviews.

src/main/java/hudson/plugins/sshslaves/verifiers/TrileadVersionSupportManager.java
```diff
@@ -0,0 +1,70 @@
...
+        } catch (ReflectiveOperationException e) {
+            // KeyAlgorithmManager doesn't exist, fall back to legacy trilead handler
+            return new LegacyTrileadVersionSupport();
+        }
+    }
+
+    private static TrileadVersionSupport createaVersion9Instance() {
+        try {
+            return (TrileadVersionSupport) Class.forName("hudson.plugins.sshslaves.verifiers.JenkinsTrilead9VersionSupport").newInstance();
+        } catch (ReflectiveOperationException e) {
```
oleg-nenashev: Any risk of ClassNotFoundException?

oleg-nenashev: I would catch all exceptions

src/main/java/hudson/plugins/sshslaves/verifiers/KnownHostsFileKeyVerificationStrategy.java
```diff
@@ -73,6 +74,16 @@ public boolean verify(SlaveComputer computer, HostKey hostKey, TaskListener list
...
     }
+
+    @Override
+    public String[] getPreferredKeyAlgorithms(SlaveComputer computer) throws IOException {
+        if (!KNOWN_HOSTS_FILE.exists()) {
+            return super.getPreferredKeyAlgorithms(computer);
+        }
+
+        KnownHosts knownHosts = new KnownHosts(KNOWN_HOSTS_FILE);
+        return knownHosts.getPreferredServerHostkeyAlgorithmOrder(((SSHLauncher) computer.getLauncher()).getHost());
```
oleg-nenashev: Unchecked cast. Please avoid it, because there is a risk somebody uses API wrong. And FindBugs will start grumbling once enabled

src/main/java/hudson/plugins/sshslaves/verifiers/TrileadVersionSupportManager.java
```diff
@@ -0,0 +1,70 @@
...
+ * @author Michael Clarke
+ * @since 1.18
+ */
+final class TrileadVersionSupportManager {
+
+    static TrileadVersionSupport getTrileadSupport() {
+        try {
+            Thread.currentThread().getContextClassLoader().loadClass("com.trilead.ssh2.signature.KeyAlgorithmManager");
+            return createaVersion9Instance();
+        } catch (ReflectiveOperationException e) {
```
oleg-nenashev: Better to catch all exceptions as well

src/main/java/hudson/plugins/sshslaves/verifiers/JenkinsTrilead9VersionSupport.java
```diff
@@ -0,0 +1,39 @@
...
+    @Override
+    public HostKey parseKey(String algorithm, byte[] keyValue) {
+        for (KeyAlgorithm<?, ?> keyAlgorithm : KeyAlgorithmManager.getSupportedAlgorithms()) {
+            try {
+                if (keyAlgorithm.getKeyFormat().equals(algorithm)) {
+                    keyAlgorithm.decodePublicKey(keyValue);
+                    return new HostKey(algorithm, keyValue);
+                }
+            } catch (IOException ex) {
+                throw new IllegalArgumentException(Messages.ManualKeyProvidedHostKeyVerifier_KeyValueDoesNotParse(algorithm), ex);
```
oleg-nenashev: This exception is not documented in TrileadVersionSupport. Instead of RunTime exceptions, I would rather suggest doing correct propagation using checked exceptions. For new APIs it is easy to do so

src/main/java/hudson/plugins/sshslaves/verifiers/TrileadVersionSupportManager.java
```diff
@@ -0,0 +1,70 @@
...
+    private static TrileadVersionSupport createaVersion9Instance() {
+        try {
+            return (TrileadVersionSupport) Class.forName("hudson.plugins.sshslaves.verifiers.JenkinsTrilead9VersionSupport").newInstance();
+        } catch (ReflectiveOperationException e) {
+            throw new IllegalArgumentException("Could not create Trilead support class", e);
+        }
+
+    }
+
+    public abstract static class TrileadVersionSupport {
```
oleg-nenashev: API of this class is not documented and IMO misused in this PR. If you want to keep this API internal for now, please mark this class and its children as `@Restricted(NoExternalUse.class)`

src/main/java/hudson/plugins/sshslaves/verifiers/TrileadVersionSupportManager.java
mc1arke: ReflectiveOperationException covers ClassNotFoundException as well as the InvocationException that would be thrown if the class could not be instantiated

src/main/java/hudson/plugins/sshslaves/verifiers/TrileadVersionSupportManager.java
mc1arke: Could you clarify what you mean by 'misused in this PR'? I can add Javadoc, but want to get the API correct. The visibility of the constructors/method in this class should prevent external use without the need for any Restriction annotations.

src/main/java/hudson/plugins/sshslaves/verifiers/TrileadVersionSupportManager.java
oleg-nenashev: Though it still does not cover other possible exceptions i n the newInstance() call. https://docs.oracle.com/javase/7/docs/api/java/lang/Class.html#newInstance()

src/main/java/hudson/plugins/sshslaves/verifiers/TrileadVersionSupportManager.java
oleg-nenashev: I am just a pretty convinced enemy of unchecked exceptions. Especially the undocumented ones. Even with disabled constructor, it is still possible to access this class e.g. by copy-pasting the `createaVersion9Instance()` code. I would keep the classes restricted if you do not want to bother about external usages now. Better safe than sorry

mc1arke:
@oleg-nenashev I think I've resolved all the points you've currently raised. Care to take another look?

paladox:
Bump. can this be merged please?

paladox:
@jglick or @oleg-nenashev hi, could you merge and do a new release of this plugin please?

mc1arke:
@stephenc could you consider reviewing, merging and releasing this please, since you seem to be the only other person with the access - unless you want me to request ssh-slaves release access myself?

MarkEWaite:
I think this plugin needs to release before the 2.60.1 LTD. Without this fix,  ssh slaves don't connect after upgrade

oleg-nenashev:
Ack. If @jglick and @olivergondza do not mind, I will integrate the ready
stuff and spin the release

stephenc:
I can review and release tomorrow. If needed sooner, someone else will need to

MarkEWaite:
I don't see any  reason to release it today. I don't know the release
schedule for 2.60.1, but I assume it isn't releasing until at least tomorrow

oleg-nenashev:
We have about 9 days before the release happens

stephenc:
@mc1arke you should also seek release permissions access.

@mc1arke can you update the release notes on the wiki page with the details of what is in the 1.18 release that I have just cut.

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -786,6 +786,7 @@ public synchronized void launch(final SlaveComputer computer, final TaskListener
             public Boolean call() throws InterruptedException {
                 Boolean rval = Boolean.FALSE;
                 try {
+                    connection.setServerHostKeyAlgorithms(sshHostKeyVerificationStrategy.getPreferredKeyAlgorithms(computer));
```
jglick: Or an IAE: [JENKINS-44832](https://issues.jenkins-ci.org/browse/JENKINS-44832)

PR #55, opened 2017-05-30
Trim whitespace from hostname

slideon:
Launching an agent failed because we had an IP with a leading space. This should fix it. Add a test as well.


slideon:
I don't know why the 4th commit still says pending, it failed due to a compilation error.

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -435,7 +435,11 @@ public SSHLauncher(String host, int port, String username, String password, Stri
     @Deprecated
     public SSHLauncher(String host, int port, String username, String password, String privatekey, String jvmOptions,
                                     String javaPath, JDKInstaller jdkInstaller, String prefixStartSlaveCmd, String suffixStartSlaveCmd) {
-        this.host = host;
+        if (host != null) {
```
oleg-nenashev: Use `hudson.Util.fixEmptyAndTrim()`. It will also convert blank lines to null, which is maybe required here

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -517,8 +521,11 @@ public SSHLauncher(String host, int port, StandardUsernameCredentials credential
     public SSHLauncher(String host, int port, StandardUsernameCredentials credentials, String jvmOptions,
                                     String javaPath, JDKInstaller jdkInstaller, String prefixStartSlaveCmd,
                                     String suffixStartSlaveCmd, Integer launchTimeoutSeconds, Integer maxNumRetries, Integer retryWaitTime, SshHostKeyVerificationStrategy sshHostKeyVerificationStrategy) {
-
-        this.host = host;
+        if (host != null) {
+            this.host = host.trim();
+        } else {
```
oleg-nenashev: same

kuisathaverat:
Also I have a plan to annotate constructors, but first I have to make a clean up https://github.com/jenkinsci/ssh-slaves-plugin/pull/91

slideon:
Thank you for finishing this. I apologize for being so inactive, I wanted to make these changes but I never found the time.

PR #59, opened 2017-06-10
Use SFTPv3Client instead of deprecated SFTPClient

darxriggs:


src/main/java/hudson/plugins/sshslaves/SFTPClient.java
```diff
@@ -1,150 +0,0 @@
...
-
-import java.io.IOException;
-import java.io.OutputStream;
-import java.io.InputStream;
-
-/**
- * TODO: moved to Jenkins core, so pick it up from there.
- * @author Kohsuke Kawaguchi
- */
-public class SFTPClient extends SFTPv3Client {
```
oleg-nenashev: Incompatible

src/main/java/hudson/plugins/sshslaves/SFTPFileSystem.java
```diff
@@ -38,9 +39,9 @@
  * @author Kohsuke Kawaguchi
  */
 class SFTPFileSystem implements FileSystem {
-    private final SFTPClient sftp;
+    private final SFTPv3Client sftp;
 
-    public SFTPFileSystem(SFTPClient sftp) {
+    public SFTPFileSystem(SFTPv3Client sftp) {
```
oleg-nenashev: Binary incompatible. Maybe makes sense to create a new SFTPv3FileSystem class instead

oleg-nenashev:
@darxriggs If you have any concerns about y feedback, feel free to respond here

darxriggs:
Thanks for the feedback. I see the point with this PR in not being backward compatible.

I think adding annotations and another class is not worth it when old code cannot be removed anyway. So I just closed the PR.

PR #61, opened 2017-06-12
[JENKINS-44830] NPE after upgrading caused by fix of JENKINS-42959

jglick:
[JENKINS-44830](https://issues.jenkins-ci.org/browse/JENKINS-44830)

Amends #54. Reproduces reported NPE.

@reviewbybees

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -272,6 +272,7 @@ public class SSHLauncher extends ComputerLauncher {
      * The verifier to use for checking the SSH key presented by the host
      * responding to the connection
      */
+    @CheckForNull
     private final SshHostKeyVerificationStrategy sshHostKeyVerificationStrategy;
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -553,10 +554,16 @@ public String getCredentialsId() {
         return credentialsId;
     }
 
+    @CheckForNull
     public SshHostKeyVerificationStrategy getSshHostKeyVerificationStrategy() {
         return sshHostKeyVerificationStrategy;
     }
 
+    @NonNull
+    SshHostKeyVerificationStrategy getSshHostKeyVerificationStrategyDefaulted() {
```
oleg-nenashev: makes sense to make public later

jglick: Only if there are some external callers—currently I do not see a need for it.

oleg-nenashev:
@reviewbybees done

jglick:
> Does this add a requirement to build on Linux that wasn't there before?

Nope, if the `docker` command is not present, the test just gets skipped (as it does on the currently configured CI job).

rtyler:
[thanks jesse](https://www.youtube.com/watch?v=ZkvMRROkbC0&t=83)

PR #63, opened 2017-06-13
[JENKINS-44832] IAE due to empty or null list of preferred key algorithms

jglick:
[JENKINS-44832](https://issues.jenkins-ci.org/browse/JENKINS-44832)

Amends #54. Not known how to reproduce, so no regression test.

@reviewbybees


oleg-nenashev:
@reviewbybees done

PR #65, opened 2017-06-30
Make "Name or service not known" errors retryable

borgstrom:
We use a dynamic DNS setup where EC2 nodes add their own DNS via userdata scripts, and entries are cleaned up on node termination.

This means that we can end up in a situation where one of our slaves DNS names can be removed and when the slave comes back online the master doesn't retry the connection.

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -159,7 +159,7 @@ public class SSHLauncher extends ComputerLauncher {
     // "No route to host", "No route to host (Host unreachable)"
     // "Premature connection close"
     private static final List<String> RECOVERABLE_FAILURES = Arrays.asList(
-            "Connection refused", "Connection reset", "Connection timed out", "No route to host", "Premature connection close"
+            "Connection refused", "Connection reset", "Connection timed out", "No route to host", "Premature connection close", "Name or service not known"
     );
 
     /**


borgstrom:
Care to explain why you do not think it's a good behavior?

borgstrom:
Specifically, what would be a negative aspect of retrying name failures?   If you've configured retries on your ssh slave and have specified a bad DNS name how does the slave retrying the connection become not a "good behavior" versus just stopping to attempt further connections?

oleg-nenashev:
The problem is that...

1) SSH agents will be infinitely trying to reconnect if the master is down
2) SSH agents will be infinitely trying to reconnect if the network is misconfigured

In both cases I would rather prefer the agent to just exit with failure

borgstrom:
I still do not understand "the problem". 

1.  If the master is down?  I don't follow your logic here. 

2.  Exactly.  I want the master to continually **retry** the connection based on the parameters I supply until the network is configured correctly again.  

Again, why do you prefer that DNS errors be hard failures?  You still haven't explained why, just what, and I understand the technical implications of this change.

borgstrom:
For example, please contrast the failure scenarios of `Name or service not known` vs. `No route to host` and tell me why the lack of a valid route entry is ok to retry but the lack of a valid host entry would be problematic.

borgstrom:
Seems futile to keep this open...

oleg-nenashev:
@borgstrom if you do not get a response, it is always possible to ping people. I cannot process all notifications I get from GitHub every day

PR #66, opened 2017-07-28
[JENKINS-29412] - Minimal Java Level should be determined dynamically

oleg-nenashev:
Now the plugin uses Jenkins core version. Also acts as a PoC for https://github.com/jenkinsci/jenkins/pull/2949

- [x] - Implement logic
- [x] - Move tests from #27 created by @daniel-beck 

https://issues.jenkins-ci.org/browse/JENKINS-29412

@reviewbybees


src/main/resources/hudson/plugins/sshslaves/Messages_es.properties
```diff
@@ -23,32 +23,32 @@
 # {0} [SSH] Remote file system root {1} does not exist. Will try to create it...
 SSHLauncher.RemoteFSDoesNotExist={0} [SSH] Directorio raiz en el sistema remoto ({1}) no existe. Intentando crearlo ...
 # {0} [SSH] Authentication successful.
-SSHLauncher.AuthenticationSuccessful={0} [SSH] Autenticaci�n v�lida
+SSHLauncher.AuthenticationSuccessful={0} [SSH] Autenticaci\u00f3n v\u00e1lida
```
MarkEWaite: Are these character transformations intentional?

src/main/java/hudson/plugins/sshslaves/JavaProvider.java
```diff
@@ -62,4 +71,31 @@
...
+            }
+        } catch(SecurityException | NoSuchMethodException | IllegalAccessException | InvocationTargetException ex) {
+            // Fallback to the default behavior, not supported yet
+        }
+        
+        // Now use the known map of the previous updates
+        final VersionNumber version = Jenkins.getVersion();
+        if (version == null) {
+            // Version cannot be determined, assume it's an old one to retain compatibility.
+            return JAVA_LEVEL_7;
```
MarkEWaite: I assume this is JAVA_LEVEL_7 rather than JAVA_LEVEL_6 because the plugin is based on Jenkins 1.625 (first LTS to mandate Java 7) and we've had one or more ugly bug reports (in the git plugin, anyway) from users trying to mix Java 7 master with Java 6 agents.

oleg-nenashev: Yes, we can forget about JAVA_LEVEL_6 in this plugin

oleg-nenashev:
I have no IDE, which saves in the original encoding format. @recena any idea what I should use?

recena:
@oleg-nenashev Let try it.

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -1251,15 +1252,17 @@ protected String checkJavaVersion(final PrintStream logger, String javaCommand,
...
+                final VersionNumber minJavaLevel = JavaProvider.getMinJavaLevel();
                 try {
                     final Number version =
                         NumberFormat.getNumberInstance(Locale.US).parse(versionStr);
-                    if(version.doubleValue() < 1.5) {
+                    //TODO: burn it with fire
+                    if(version.doubleValue() < Double.parseDouble("1."+minJavaLevel)) {
                         throw new IOException(Messages
-                                .SSHLauncher_NoJavaFound(line));
+                                .SSHLauncher_NoJavaFound(line, minJavaLevel.toString()));
```
jglick: `.toString()` gratuitous

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -1251,15 +1252,17 @@ protected String checkJavaVersion(final PrintStream logger, String javaCommand,
                         getTimestamp(), javaCommand, versionStr));
 
                 // parse as a number and we should be OK as all we care about is up through the first dot.
+                final VersionNumber minJavaLevel = JavaProvider.getMinJavaLevel();
                 try {
                     final Number version =
                         NumberFormat.getNumberInstance(Locale.US).parse(versionStr);
-                    if(version.doubleValue() < 1.5) {
+                    //TODO: burn it with fire
+                    if(version.doubleValue() < Double.parseDouble("1."+minJavaLevel)) {
```
jglick: Can we not use `VersionNumber.compare` on `versionStr`?

src/main/resources/hudson/plugins/sshslaves/Messages_es.properties
```diff
@@ -23,32 +23,32 @@
 # {0} [SSH] Remote file system root {1} does not exist. Will try to create it...
 SSHLauncher.RemoteFSDoesNotExist={0} [SSH] Directorio raiz en el sistema remoto ({1}) no existe. Intentando crearlo ...
 # {0} [SSH] Authentication successful.
-SSHLauncher.AuthenticationSuccessful={0} [SSH] Autenticaci�n v�lida
+SSHLauncher.AuthenticationSuccessful={0} [SSH] Autenticaci�n v�lida
```
jglick: Looks like you corrupted this file.

Anyway you did not touch all the translations defining `SSHLauncher.NoJavaFound`. Perhaps better to just leave the translations alone, and define a new key.

oleg-nenashev: i didn't corrupt it, works in my IDE (c). Can get better encoding in Netbeans. New key/entry deletion won't help since I would have to save the file anyway. Ping @recena who was going to take a look

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
oleg-nenashev: Yes, we could. Such approach is better then the original one I'd guess. Will need to write tests though

oleg-nenashev:
@recena ping

oleg-nenashev:
@jglick done

src/main/resources/hudson/plugins/sshslaves/Messages.properties
```diff
@@ -10,7 +10,8 @@ SSHLauncher.ConnectionClosed={0} [SSH] Connection closed.
 SSHLauncher.ErrorWhileClosingConnection=Exception thrown while closing connection.
 SSHLauncher.AbortedDuringConnectionOpen=Slave start aborted.
 SSHLauncher.FailedToDetectEnvironment=Failed to detect the environment for automatic JDK installation. Please report this to jenkinsci-users@googlegroups.com: {0}
-SSHLauncher.NoJavaFound=Java version {0} was found but 1.5 or later is needed.
+SSHLauncher.NoJavaFound=Java version {0} was found but 1.7 or later is needed.
```
jglick: Delete this key—it is obsolete and translators need to know that.

oleg-nenashev: I prefer to keep it since I cannot modify the Spanish file. Once I find out how to do it, I will remove both

oleg-nenashev:
@reviewbybees done
@daniel-beck any concerns?

src/main/resources/hudson/plugins/sshslaves/Messages.properties
jglick: > I cannot modify the Spanish file

Exactly my point: once you delete the English key, translators should be able to see that they have stale keys in their translations and delete them on their own.

src/main/java/hudson/plugins/sshslaves/JavaProvider.java
```diff
@@ -62,4 +71,35 @@
...
+    /**
+     * Gets minimal required Java version.
+     * 
+     * @return Minimal Java version required on the master and agent side.
+     *         It will be {@link #JAVA_LEVEL_7} if the core version cannot be determined due to whatever reason.
+     * @since TODO
+     * 
+     */
+    @Nonnull
+    public static VersionNumber getMinJavaLevel() {
```
jtnord: I don't get it.  The best practice is to run the same major version of java on both the master and agents, so why not just check what is running and use that.  
If you need to know the minimum then you do not need an API in core for that - you can do it today without an API by finding out the version of the classfile of `Jenkins`

daniel-beck: Except for the case of plugins that require newer Java than core, and are running on the master.

PR #68, opened 2017-08-23
[JENKINS-44893] Fix Trilead support class instantiation

gmshake:
Issue link: [https://issues.jenkins-ci.org/browse/JENKINS-44893](https://issues.jenkins-ci.org/browse/JENKINS-44893)

Since the JenkinsTrilead9VersionSupport is within the same classpath as TrileadVersionSupportManager, we can safely instantiate it without worrying class loading issues.

### DIFF
--- a/src/test/java/hudson/plugins/sshslaves/verifiers/TrileadVersionSupportManagerTest.java
+++ b/src/test/java/hudson/plugins/sshslaves/verifiers/TrileadVersionSupportManagerTest.java
@@ -11,6 +11,7 @@ import java.util.List;
 import java.util.Set;
 
 import org.junit.Test;
+import org.jvnet.hudson.test.Issue;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotEquals;
@@ -31,6 +32,7 @@ public class TrileadVersionSupportManagerTest {
     }
 
     @Test
+    @Issue("JENKINS-44893")
     public void testLegacyInstanceWithLinkageError() {
         BlockingClassloader classloader = newBlockingClassloader();
         classloader.inspectPackage("com.trilead.ssh2.signature");
@@ -46,6 +48,7 @@ public class TrileadVersionSupportManagerTest {
     }
 
     @Test
+    @Issue("JENKINS-44893")
     public void testCurrentInstanceWithIsolatedClassLoader() {
         BlockingClassloader classloader = newBlockingClassloader();
         Object trileadSupport = invokeGetTrileadSupport(classloader);
--- a/src/test/java/hudson/plugins/sshslaves/verifiers/TrileadVersionSupportManagerTest.java
+++ b/src/test/java/hudson/plugins/sshslaves/verifiers/TrileadVersionSupportManagerTest.java
@@ -31,6 +31,16 @@ public class TrileadVersionSupportManagerTest {
     }
 
     @Test
+    public void testLegacyInstanceWithLinkageError() {
+        BlockingClassloader classloader = newBlockingClassloader();
+        classloader.inspectPackage("com.trilead.ssh2.signature");
+        classloader.block("com.trilead.ssh2.signature.KeyAlgorithm");
+
+        Object trileadSupport = invokeGetTrileadSupport(classloader);
+        assertEquals("hudson.plugins.sshslaves.verifiers.TrileadVersionSupportManager$LegacyTrileadVersionSupport", trileadSupport.getClass().getName());
+    }
+
+    @Test
     public void testCurrentInstance() {
         assertEquals(JenkinsTrilead9VersionSupport.class, TrileadVersionSupportManager.getTrileadSupport().getClass());
     }
--- a/src/test/java/hudson/plugins/sshslaves/verifiers/TrileadVersionSupportManagerTest.java
+++ b/src/test/java/hudson/plugins/sshslaves/verifiers/TrileadVersionSupportManagerTest.java
@@ -1,24 +1,33 @@
 package hudson.plugins.sshslaves.verifiers;
 
+import java.io.IOException;
+import java.lang.reflect.Method;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.util.ArrayList;
+import java.util.Enumeration;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


gmshake:
Sorry for late response.
To be clear, createVersion9Instance() is protected by isAfterTrilead8() in TrileadVersionSupportManager line 31. If I'm right, isAfterTrilead8() checks dependencies of JenkinsTrilead9VersionSupport at runtime. So anyone who doesn't have an appropriate version of Trilead will NOT run to createVersion9Instance().

The real problem is that, Thread.currentThread().getContextClassLoader() is not always the classloader of caller, aka TrileadVersionSupportManager. Jenkins plugin has its own classloader. I'd prefer to rewrite the related testcase, any idea?

gmshake:
@mc1arke Any progress on this issue?

mc1arke:
Your code change adds an import (so non reflective reference) to JenkinsTrilead9VersionSupport, which has an import on Trilead 9 version classes, so TrileadVersionSupport manager would fail to load if Trilead 9 classes weren't on the classpath. I'm happy for you to change the referenced classloader if needed, but this change as it stands would just cause issues for anyone with older core versions.

gmshake:
Uhm, what about this one

```
private static TrileadVersionSupport createVersion9Instance() throws ReflectiveOperationException {
    return (TrileadVersionSupport) TrileadVersionSupportManager.class.getClassLoader()
                .loadClass("hudson.plugins.sshslaves.verifiers.JenkinsTrilead9VersionSupport").newInstance();
}
```

or this simpler one
```
private static TrileadVersionSupport createVersion9Instance() throws ReflectiveOperationException {
    return (TrileadVersionSupport) Class.forName("hudson.plugins.sshslaves.verifiers.JenkinsTrilead9VersionSupport").newInstance();
}
```

Both tested and works.

gmshake:
@mc1arke Barrier isAfterTrilead8() is rewritten, and test cases are updated as respect :)

gmshake:
Ping @mc1arke

gmshake:
@reviewbybees Done.

gmshake:
@daniel-beck Could you please merge it?

src/test/java/hudson/plugins/sshslaves/verifiers/TrileadVersionSupportManagerTest.java
```diff
@@ -1,44 +1,144 @@
...
-        }
+        BlockingClassloader classloader = newBlockingClassloader();
+        classloader.block("com.trilead.ssh2.signature.KeyAlgorithmManager");
+
+        Object trileadSupport = invokeGetTrileadSupport(classloader);
+        assertEquals("hudson.plugins.sshslaves.verifiers.TrileadVersionSupportManager$LegacyTrileadVersionSupport", trileadSupport.getClass().getName());
+    }
+
+    @Test
+    public void testLegacyInstanceWithLinkageError() {
```
oleg-nenashev: Nice2have: `@Issue("JENKINS-44893")` annotation

src/test/java/hudson/plugins/sshslaves/verifiers/TrileadVersionSupportManagerTest.java
```diff
@@ -1,44 +1,144 @@
...
+        assertEquals("hudson.plugins.sshslaves.verifiers.TrileadVersionSupportManager$LegacyTrileadVersionSupport", trileadSupport.getClass().getName());
     }
 
     @Test
     public void testCurrentInstance() {
         assertEquals(JenkinsTrilead9VersionSupport.class, TrileadVersionSupportManager.getTrileadSupport().getClass());
     }
 
+    @Test
+    public void testCurrentInstanceWithIsolatedClassLoader() {
```
oleg-nenashev: same

gmshake: Done :)

src/test/java/hudson/plugins/sshslaves/verifiers/TrileadVersionSupportManagerTest.java
gmshake: done

src/test/java/hudson/plugins/sshslaves/verifiers/TrileadVersionSupportManagerTest.java
```diff
@@ -1,44 +1,147 @@
...
+    }
+
+    private static BlockingClassloader newBlockingClassloader() {
+        BlockingClassloader classloader = new BlockingClassloader(TrileadVersionSupportManagerTest.class.getClassLoader());
+        classloader.inspectPackage("hudson.plugins.sshslaves");
+        return classloader;
+    }
+
+
+    private static class BlockingClassloader extends URLClassLoader {
```
jglick: Not the standard `MaskingClassLoader`?

gmshake: I did not know there is a 'MaskingClassLoader', I'll look at it 👍

gmshake: @jglick 
I tried MaskingClassLoader but did not get it right. Correct me if I'm wrong.

The MaskingClassLoader uses parent-first strategy when loading class, thus will unable to 'mask' classes referenced by that is loaded with parent classloader. Referrer https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-5.html#jvms-5.3.

The BlockingClassloader uses parent-last strategy. When loading 'inspected' classes/packages, it becomes a 'defining' classloader, and will be able to 'block' any classes as will, including java.lang.Object class, realistic to create an isolated test environment.

In this test case, I'd prefer BlockingClassloader :)

oleg-nenashev: The justification LGTM, just a test code anyway

gmshake: ping @jglick

oleg-nenashev:
I tried the change on 1.651.3 and 2.73.2, it seems to work well. 🚢 🇮🇹 , I'd guess. @jglick are you fine with the response above?

PR #75, opened 2017-11-21
[JENKINS-19465] - Prevent hanging of launcher when we disconnect the agent before the full launch operation execution.

oleg-nenashev:
There is hanging we see: 

```
"SSHLauncher.launch for 'MyAgent' node [#1]" #559943 prio=5 os_prio=0 tid=0x00007f9b0d994000 nid=0x4ce5 in Object.wait() [0x00007f9ab01b5000]
   java.lang.Thread.State: WAITING (on object monitor)
	at java.lang.Object.wait(Native Method)
	at java.lang.Thread.join(Thread.java:1252)
	- locked <0x000000059e20bea8> (a com.trilead.ssh2.Connection$PumpThread)
	at java.lang.Thread.join(Thread.java:1326)
	at com.trilead.ssh2.Connection.exec(Connection.java:1574)
	at hudson.plugins.sshslaves.SSHLauncher.checkJavaVersion(SSHLauncher.java:1208)
	at hudson.plugins.sshslaves.SSHLauncher.resolveJava(SSHLauncher.java:902)
	at hudson.plugins.sshslaves.SSHLauncher$2.call(SSHLauncher.java:808)
	at hudson.plugins.sshslaves.SSHLauncher$2.call(SSHLauncher.java:792)
```

Both agents may be a problem, because we see hundreds of afterDisconnect() handlers waiting for the lock.

```
"Computer.threadPoolForRemoting [#49225]" #1900616 daemon prio=5 os_prio=0 tid=0x00007f9b5482f800 nid=0x3312 waiting for monitor entry [0x00007f9a0fe40000]
   java.lang.Thread.State: BLOCKED (on object monitor)
	at hudson.plugins.sshslaves.SSHLauncher.afterDisconnect(SSHLauncher.java:1336)
	- waiting to lock <0x000000058143f168> (a hudson.plugins.sshslaves.SSHLauncher)
	at hudson.slaves.SlaveComputer$3.run(SlaveComputer.java:633)
	at jenkins.util.ContextResettingExecutorService$1.run(ContextResettingExecutorService.java:28)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:748)
```

If the `launch()` operation hangs, the `afterDisconnect()` command will not interrupt it because it runs in a separate executor service. And, if you have no "launchTimeout" property set, it will hang forever (?).

https://issues.jenkins-ci.org/browse/JENKINS-19465

@reviewbybees

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -1371,7 +1371,7 @@ public class SSHLauncher extends ComputerLauncher {
         tearDownConnection(slaveComputer, listener);
     }
 
-    private synchronized void tearDownConnection(@Nonnull SlaveComputer slaveComputer, @Nonnull TaskListener listener) {
+    private synchronized void tearDownConnection(@Nonnull SlaveComputer slaveComputer, final @Nonnull TaskListener listener) {
         if (connection != null) {
             boolean connectionLost = reportTransportLoss(connection, listener);
             if (session != null) {
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -849,13 +849,14 @@ public class SSHLauncher extends ComputerLauncher {
         try {
             long time = System.currentTimeMillis();
             List<Future<Boolean>> results;
-            if (launcherExecutorService == null) {
+            final ExecutorService srv = launcherExecutorService;
+            if (srv == null) {
                 throw new IllegalStateException("Launcher Executor Service should be always non-null here, because the task allocates and closes service on its own");
             }
             if (this.getLaunchTimeoutMillis() > 0) {
-                results = launcherExecutorService.invokeAll(callables, this.getLaunchTimeoutMillis(), TimeUnit.MILLISECONDS);
+                results = srv.invokeAll(callables, this.getLaunchTimeoutMillis(), TimeUnit.MILLISECONDS);
             } else {
-                results = launcherExecutorService.invokeAll(callables);
+                results = srv.invokeAll(callables);
             }
             long duration = System.currentTimeMillis() - time;
             Boolean res;
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -237,7 +237,7 @@ public class SSHLauncher extends ComputerLauncher {
     /**
      * SSH connection to the slave.
      */
-    private transient Connection connection;
+    private transient volatile Connection connection;
 
     /**
      * The session inside {@link #connection} that controls the slave process.
@@ -268,6 +268,12 @@ public class SSHLauncher extends ComputerLauncher {
      * Field retryWaitTime.
      */
     public final Integer retryWaitTime;
+
+    /**
+     * Keeps executor service for the async launch operation.
+     */
+    @javax.annotation.CheckForNull
+    private transient volatile ExecutorService launcherExecutorService;
 
     /**
      * The verifier to use for checking the SSH key presented by the host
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -862,12 +869,16 @@ public Boolean call() throws InterruptedException {
                 System.out.println(Messages.SSHLauncher_LaunchCompletedDuration(getTimestamp(),
                         nodeName, host, duration));
             }
-            executorService.shutdown();
```
oleg-nenashev: It was not shutting down the executor service when we were interrupting the launch :(

oleg-nenashev:
Hmm, FindBugs passes with assert on my machine 🤷‍♂️

oleg-nenashev:
Diff without formatting: https://github.com/jenkinsci/ssh-slaves-plugin/pull/75/files?w=1

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -862,12 +874,16 @@ public Boolean call() throws InterruptedException {
...
         } catch (InterruptedException e) {
             System.out.println(Messages.SSHLauncher_LaunchFailed(getTimestamp(),
                     nodeName, host));
+        } finally {
+            ExecutorService srv = launcherExecutorService;
+            if (srv != null) {
+                srv.shutdown();
+                launcherExecutorService = null;
```
jglick: Maybe swap these two lines to avoid a race condition?

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -1338,78 +1354,91 @@ private boolean isRecoverable(String message) {
...
+        }
+
+        ExecutorService srv = launcherExecutorService;
+        if (srv != null) {
+            // If the service is still running, shut it down and interrupt the operations if any
+            srv.shutdown();
+        }
+
+        synchronized (this) {
+            if (connection != null) {
```
jglick: Would prefer less reindentation. Even with the `?w=1` diff, there could be issues backporting, merging, blaming, etc.

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
oleg-nenashev: makes sense assuming that `srv.shutdown()` never throws RuntimeException. It should not according to https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html#shutdown(). Maybe it also makes sense to use https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html#shutdownNow(), we can abort this executor service right away

jglick:
@reviewbybees done

oleg-nenashev:
WiP, actually

oleg-nenashev:
@reviewbybees needs re-reviews

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -1338,7 +1356,22 @@ private boolean isRecoverable(String message) {
...
+        ExecutorService srv = launcherExecutorService;
+        if (srv != null) {
+            // If the service is still running, shut it down and interrupt the operations if any
+            srv.shutdown();
+        }
+
+        tearDownConnection(slaveComputer, listener);
+    }
+
+    private synchronized void tearDownConnection(@Nonnull SlaveComputer slaveComputer, final @Nonnull TaskListener listener) {
```
jglick: Why are you introducing this method that is called only once? Could remove ~4 diff lines.

oleg-nenashev: Otherwise I would have to use `synchronized(this) {` which would tabulate the code. And that's what you asked to avoid in https://github.com/jenkinsci/ssh-slaves-plugin/pull/75#discussion_r152390592

PR #76, opened 2017-11-28
[JENKINS-48260] Upgrade parent POM and fix PCT

fcojfernandez:
See [JENKINS-48260](https://issues.jenkins-ci.org/browse/JENKINS-48260)

## Features:

- Upgrade to `parent-pom:3.0` and update dependencies
- Upgrading baseline to `jenkins.version:2.7` to prevent `IndexOutOfBoundsException` when the ExtensionList is recovered.
- `TrustHostKeyActionTest` updated with new dependency `sshd-core:1.6.0`
- `SSHLauncherTest` updated to check if Java 7 is supported depending on `jenkins.version`

@reviewbybees spec. @stephenc 
@olamy 
@mc1arke

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.11</version>
+    <version>3.0</version>
   </parent>
 
   <artifactId>ssh-slaves</artifactId>
@@ -24,9 +24,8 @@
   </licenses>
 
   <properties>
-    <jenkins.version>1.625</jenkins.version>
+    <jenkins.version>2.7.1</jenkins.version>
     <java.level>7</java.level>
-    <jenkins-test-harness.version>2.18</jenkins-test-harness.version>
   </properties>
 
   <developers>
@@ -92,6 +91,24 @@
       <groupId>org.jenkins-ci.test</groupId>
       <artifactId>docker-fixtures</artifactId>
       <version>1.0</version>
+      <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <!-- provided by jenkins-core -->
+          <groupId>com.google.inject</groupId>
+          <artifactId>guice</artifactId>
+        </exclusion>
+        <exclusion>
+          <!-- provided by jenkins-core -->
+          <groupId>org.jenkins-ci</groupId>
+          <artifactId>annotation-indexer</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.sshd</groupId>
+      <artifactId>sshd-core</artifactId>
+      <version>1.6.0</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -90,20 +90,8 @@
     <dependency>
       <groupId>org.jenkins-ci.test</groupId>
       <artifactId>docker-fixtures</artifactId>
-      <version>1.0</version>
+      <version>1.4</version>
       <scope>test</scope>
-      <exclusions>
-        <exclusion>
-          <!-- provided by jenkins-core -->
-          <groupId>com.google.inject</groupId>
-          <artifactId>guice</artifactId>
-        </exclusion>
-        <exclusion>
-          <!-- provided by jenkins-core -->
-          <groupId>org.jenkins-ci</groupId>
-          <artifactId>annotation-indexer</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
   </dependencies>
 
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
   </licenses>
 
   <properties>
-    <jenkins.version>2.7.1</jenkins.version>
+    <jenkins.version>1.625</jenkins.version>
     <java.level>7</java.level>
   </properties>
 
@@ -104,12 +104,6 @@
           <artifactId>annotation-indexer</artifactId>
         </exclusion>
       </exclusions>
-    </dependency>
-    <dependency>
-      <groupId>org.apache.sshd</groupId>
-      <artifactId>sshd-core</artifactId>
-      <version>1.6.0</version>
-      <scope>test</scope>
     </dependency>
   </dependencies>
 
--- a/src/test/java/hudson/plugins/sshslaves/SSHLauncherTest.java
+++ b/src/test/java/hudson/plugins/sshslaves/SSHLauncherTest.java
@@ -52,6 +52,9 @@ import hudson.model.Slave;
 import hudson.slaves.DumbSlave;
 import hudson.slaves.RetentionStrategy;
 import java.util.concurrent.ExecutionException;
+
+import hudson.util.VersionNumber;
+import jenkins.model.Jenkins;
 import org.jenkinsci.test.acceptance.docker.DockerRule;
 import org.jenkinsci.test.acceptance.docker.fixtures.JavaContainer;
 import org.junit.Assert;
@@ -74,6 +77,8 @@ import org.jvnet.hudson.test.Issue;
 
 public class SSHLauncherTest {
 
+    private static final VersionNumber JAVA_8_MINIMAL_SINCE = new VersionNumber("2.54");
+
     @ClassRule
     public static BuildWatcher buildWatcher = new BuildWatcher();
 
@@ -85,7 +90,13 @@ public class SSHLauncherTest {
 
     @Test
     public void checkJavaVersionOpenJDK7NetBSD() throws Exception {
-        assertTrue("OpenJDK7 on NetBSD should be supported", checkSupported("openjdk-7-netbsd.version"));
+        final VersionNumber jenkinsVersion = Jenkins.getVersion();
+
+        if(jenkinsVersion.isOlderThan(JAVA_8_MINIMAL_SINCE)) {
+            assertTrue("OpenJDK7 on NetBSD should be supported", checkSupported("openjdk-7-netbsd.version"));
+        } else {
+            assertNotSupported("openjdk-7-netbsd.version");
+        }
     }
 
     @Test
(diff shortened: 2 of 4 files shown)


pom.xml
```diff
@@ -93,6 +92,24 @@
       <artifactId>docker-fixtures</artifactId>
       <version>1.0</version>
       <scope>test</scope>
+      <exclusions>
```
jglick: [changelog](https://github.com/jenkinsci/docker-fixtures/#11-2017-oct-12)

pom.xml
```diff
@@ -24,9 +24,8 @@
   </licenses>
 
   <properties>
-    <jenkins.version>1.625</jenkins.version>
+    <jenkins.version>2.7.1</jenkins.version>
```
oleg-nenashev: What is the justification to bump the version so far?

pom.xml
```diff
@@ -93,6 +92,24 @@
...
+          <groupId>org.jenkins-ci</groupId>
+          <artifactId>annotation-indexer</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.sshd</groupId>
+      <artifactId>sshd-core</artifactId>
+      <version>1.6.0</version>
+      <scope>test</scope>
```
oleg-nenashev: I do not get that. Jenkins core ships its own version. By using it you make the test environment different from the real one. Why?

src/test/java/hudson/plugins/sshslaves/SSHLauncherTest.java
```diff
@@ -74,6 +77,8 @@
 
 public class SSHLauncherTest {
 
+    private static final VersionNumber JAVA_8_MINIMAL_SINCE = new VersionNumber("2.54");
```
oleg-nenashev: Please avoid copy-pasting from https://github.com/jenkinsci/ssh-slaves-plugin/blob/f0634f15bdc10a67cc364a191875dc258d127334/src/main/java/hudson/plugins/sshslaves/JavaProvider.java#L45-L47 . If you need it, it is better to extend the JavaProvider API if needed. 🐛

pom.xml
fcojfernandez: @jglick AFIU, you're proposing to upgrade the docker-fixtures version instead of exclude dependencies. Currently it is not possible due to JavaContainer can't be built.
I've raised this [JIRA issue](https://issues.jenkins-ci.org/browse/JENKINS-48310) and commited this [PR](https://github.com/jenkinsci/docker-fixtures/pull/11) to fix it. I think a new version will have to be released.

pom.xml
fcojfernandez: This comment and the next one are related, so I'll try to give a common answer to be clearer.

`ssh-slaves` is using the inherited dependency (with test scope) `sshd-core` from the `jenkins-core`. That makes the plugin to be extremely dependent on the `jenkins-core` version. Older versions had a dependency with `sshd-core:0.x`, while `jenkins-core:2.x` have dependency with `sshd-core:1.x`. Many packages and classes have been refactored and moved, so tests would fail depending on the environment we execute them:
- They will succeed in a stand-alone execution or in an environment based on `jenkins-core` versions lower than 2.0.
- They will fail in an environment based on `jenkins-core` version 2.0 or upper.
Trying to avoid those failures, I propose to fix to `ssh-core:1.6.0` (The dependency for newer `jenkins-core` version) and updated the tests. IMO, this will create an actual isolated test environment.
The problem is SSHD module in jenkins-core is using the class `KeyPairProvider` in `sshd-core`, that is one of the refactored classes. Keeping the `jenkins-core:1.625` causes an exception when trying to get the ExtensionList, so it should be upgraded. I have proposed 2.7.1 just because it is a LTS:

`org.jvnet.hudson.reactor.ReactorException: java.lang.Error: java.lang.reflect.InvocationTargetException
        at org.jvnet.hudson.reactor.Reactor.execute(Reactor.java:269)
        at jenkins.InitReactorRunner.run(InitReactorRunner.java:44)
        at jenkins.model.Jenkins.executeReactor(Jenkins.java:916)
        at jenkins.model.Jenkins.<init>(Jenkins.java:815)
        at hudson.model.Hudson.<init>(Hudson.java:85)
        at org.jvnet.hudson.test.JenkinsRule.newHudson(JenkinsRule.java:611)
        at org.jvnet.hudson.test.JenkinsRule.before(JenkinsRule.java:384)
        at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:537)
        at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
        at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
        at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.Error: java.lang.reflect.InvocationTargetException
        at hudson.init.TaskMethodFinder.invoke(TaskMethodFinder.java:110)
        at hudson.init.TaskMethodFinder$TaskImpl.run(TaskMethodFinder.java:176)
        at org.jvnet.hudson.reactor.Reactor.runTask(Reactor.java:282)
        at jenkins.model.Jenkins$7.runTask(Jenkins.java:905)
        at org.jvnet.hudson.reactor.Reactor$2.run(Reactor.java:210)
        at org.jvnet.hudson.reactor.Reactor$Node.run(Reactor.java:117)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        ... 1 more
Caused by: java.lang.reflect.InvocationTargetException
        at sun.reflect.GeneratedMethodAccessor36.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at hudson.init.TaskMethodFinder.invoke(TaskMethodFinder.java:106)
        ... 8 more
Caused by: java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
        at java.util.ArrayList.rangeCheck(ArrayList.java:653)
        at java.util.ArrayList.get(ArrayList.java:429)
        at hudson.ExtensionList.get(ExtensionList.java:171)
        at hudson.PluginManager$PluginUpdateMonitor.getInstance(PluginManager.java:1258)
        at hudson.maven.PluginImpl.init(PluginImpl.java:58)
        ... 12 more`

I hope to make myself understood.

pom.xml
fcojfernandez: [See previous comment](https://github.com/jenkinsci/ssh-slaves-plugin/pull/76#discussion_r154013397)

oleg-nenashev:
@fcojfernandez The impact of SSHD Core 1.x incompatibility is "just" one test with 2 class instatinations, right? What if we just put these class instantiations to the Reflection code? Would it help PCT to keep running on the current and old core version?

I am not against bumping to 2.7.2 per-se, but I am against adding such tweaks to all tests just to get 1 test running. And it's too early to bump to the 2.73.x baseline with new SSHD Core.

fcojfernandez:
@oleg-nenashev Although I have no strong opinion about using reflection in this case, I'm not totally confortable with the fact that in the future, if sshd-core version used in jenkins-core changes again and there are new refactors this test will have to be updated again. However, your proposal is less intrusive than mine, so I've decided to follow your indications.
What do you think now?

fcojfernandez:
@oleg-nenashev @jglick what do you think now?

src/test/java/hudson/plugins/sshslaves/verifiers/TrustHostKeyActionTest.java
```diff
@@ -93,29 +85,33 @@ public void testSubmitNotAuthorised() throws Exception {
...
+            invoke(server, "setPort", new Class[] {Integer.TYPE}, new Object[] {port});
+            invoke(server, "setKeyPairProvider", new Class[] {keyPairProviderClass}, new Object[] {provider});
+            invoke(server, "setUserAuthFactories", new Class[] {List.class}, new Object[] {Arrays.asList(factory)});
+            invoke(server, "setCommandFactory", new Class[] {commandFactoryClass}, new Object[] {commandFactory});
+            invoke(server, "setPasswordAuthenticator", new Class[] {commandAuthenticatorClass}, new Object[] {authenticator});
+
+            invoke(server, "start", null, null);
+        } catch (ClassNotFoundException | NoSuchMethodException | InvocationTargetException | IllegalAccessException | InstantiationException | IllegalArgumentException e) {
+            e.printStackTrace();
+            throw new AssertionError("Check sshd-core version");
```
jglick: Why not just let these be thrown up?

src/test/java/hudson/plugins/sshslaves/verifiers/TrustHostKeyActionTest.java
```diff
@@ -144,5 +140,113 @@ public boolean authenticate(String username, String password, ServerSession sess
...
     }
-    
+
+    private Object newSshServer() throws ClassNotFoundException, NoSuchMethodException, InvocationTargetException, IllegalAccessException {
+        Object server = null;
+        Class serverClass;
+        try {
+            serverClass = Class.forName("org.apache.sshd.SshServer");
+        } catch (ClassNotFoundException e) {
+            serverClass = Class.forName("org.apache.sshd.server.SshServer");
```
jglick: You need to edit `Jenkinsfile` to say, e.g.

```groovy
buildPlugin(jenkinsVersions: [null, '2.89.1'])
```

so that CI actually tests both branches.

src/test/java/hudson/plugins/sshslaves/verifiers/TrustHostKeyActionTest.java
fcojfernandez: IMO, this message is more easy to understand than an InvocationTargetException (e.g.) and gives the root cause of a possible error. However, I agree with you about throwing the original exception to have the detailed error so I will change it into 
`throw new AssertionError("Check sshd-core version", e);`

src/test/java/hudson/plugins/sshslaves/verifiers/TrustHostKeyActionTest.java
fcojfernandez: OK! But as version '2.73.1' shares the same sshd-core library than '2.89.1', I think it will be better to use it

oleg-nenashev:
Will include it into the 1.24 release so that others have more time to review.
CC @jglick

PR #78, opened 2017-12-18
[JENKINS-48613] - Hotfix: Prevent piling up of tearDownConnection() calls if they take long

oleg-nenashev:
It extends my JENKINS-19465 fix in #75 and makes the hack even more hackish.
Ideally Jenkins Core needs to be updated to somehow prevent /queue such duplicated calls in the Executor service. Should be done on a ComputerListener level or so. I will create tickets for that later

@reviewbybees @jglick

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -125,6 +125,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
 
@@ -240,6 +241,12 @@ public class SSHLauncher extends ComputerLauncher {
      * SSH connection to the slave.
      */
     private transient volatile Connection connection;
+
+    /**
+     * Indicates that the {@link #tearDownConnection(SlaveComputer, TaskListener)} is in progress.
+     * It is used in {@link #afterDisconnect(SlaveComputer, TaskListener)} to avoid multiple parallel calls.
+     */
+    private transient volatile boolean tearingDownConnection;
 
     /**
      * The session inside {@link #connection} that controls the slave process.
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -1368,11 +1375,24 @@ public void afterDisconnect(SlaveComputer slaveComputer, final TaskListener list
             srv.shutdown();
         }
 
+        if (tearingDownConnection) {
```
jglick: Could we simply move the call to `tearDownConnection` into a background thread so that we do not block `afterDisconnect`?

jglick:
BTW

    Possible null pointer dereference in hudson.plugins.sshslaves.SSHLauncher$DescriptorImpl.getHelpFile(String) due to return value of called method [hudson.plugins.sshslaves.SSHLauncher$DescriptorImpl, hudson.plugins.sshslaves.SSHLauncher$DescriptorImpl] Dereferenced at SSHLauncher.java:[line 1620]Known null at SSHLauncher.java:[line 1620] NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE

jglick:
(Broken in `master`, but somehow #76 was clean; not sure how that happened.)

jglick:
Ah, right:

    Loading trusted files from base branch master at a413f56449b3fb602a870929fe08cdad3d0d63bf rather than 77ed6569ea7e72a87029709bc288efcbe8a8c7a5

Was submitted by a nontrusted contributor, so the actual `Jenkinsfile` modifications were not checked. We should really make this fatal I suppose. Will fix this case.

jglick:
Merge with #79 to retest.

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
oleg-nenashev: That's what I would like to do (see my comment in the PR description).

oleg-nenashev:
I have created [JENKINS-48616](https://issues.jenkins-ci.org/browse/JENKINS-48616), [JENKINS-48617](https://issues.jenkins-ci.org/browse/JENKINS-48617) and [JENKINS-48618] for timeout follow-ups

PR #79, opened 2017-12-18
FindBugs

jglick:
Fixes a problem in #76 which was not caught since @fcojfernandez has no write permission to the repository, and so his change to build against 2.73.x was never actually tested.

@reviewbybees


jglick:
@reviewbybees done

PR #80, opened 2017-12-18
[JENKINS-48616] - Pass launch timeout as Connection and KEx timeout to TrileadSSH

oleg-nenashev:
https://issues.jenkins-ci.org/browse/JENKINS-48616

It does not make much difference without [JENKINS-48617](https://issues.jenkins-ci.org/browse/JENKINS-48617) and [JENKINS-48618](https://issues.jenkins-ci.org/browse/JENKINS-48618), but OTOH it is a low-hanging fruit.

@reviewbybees @jglick

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -1299,6 +1299,10 @@ public class SSHLauncher extends ComputerLauncher {
         int maxNumRetries = this.maxNumRetries == null || this.maxNumRetries < 0 ? 0 : this.maxNumRetries;
         for (int i = 0; i <= maxNumRetries; i++) {
             try {
+                // We pass launch timeout so that the connection will be able to abort once it reaches the timeout
+                // It is a poor man's logic, but it should cause termination if the connection goes strongly beyond the timeout
+                //TODO: JENKINS-48617 and JENKINS-48618 need to be implemented to make it fully robust
+                int launchTimeoutMillis = (int)getLaunchTimeoutMillis();
                 connection.connect(new ServerHostKeyVerifier() {
 
                     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -1299,6 +1299,10 @@ protected void openConnection(final TaskListener listener, final SlaveComputer c
         int maxNumRetries = this.maxNumRetries == null || this.maxNumRetries < 0 ? 0 : this.maxNumRetries;
         for (int i = 0; i <= maxNumRetries; i++) {
             try {
+                // We pass launch timeout so that the connection will be able to abort once it reaches the timeout
+                // It is a poor man's logic, but it should cause termination if the connection goes strongly beyond the timeout
+                //TODO: JENKINS-48617 and JENKINS-48618 need to be implemented to make it fully robust
+                int launchTimeoutMillis = (int)getLaunchTimeoutMillis();
```
jglick: Possible integer overflow? I suppose it is unlikely that we would want to wait 2³¹ms, but…

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -1308,7 +1312,7 @@ public boolean verifyServerHostKey(String hostname, int port, String serverHostK
 
                         return getSshHostKeyVerificationStrategyDefaulted().verify(computer, key, listener);
                     }
-                });
+                }, launchTimeoutMillis, 0 /*read timeout - JENKINS-48618*/, launchTimeoutMillis);
```
jglick: Does this behave as before when `launchTimeoutMillis == 0`?

oleg-nenashev: Yes, that's what the original method passes to this one

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
oleg-nenashev: Yeah, maybe I should add handling for that. But yeah, it it a bit paranoid in this case.
It's roughly 24.8 days 🤷‍♂️

PR #82, opened 2018-02-15
don’t lookup systemCredentials to eventually store credentialsId which was initial constructor parameter

ndeloof:
SSHLauncher do receive `credentialsId` but then retrieve matching `Credentials` instance, then stores `credentials.getI()`. Proposed change make it just simpler by storing ID

initial reason doing so :
Running Configuration-as-Code we inject credentialsId during init, while the credentials store might not be setup yet so it can resolve it to an actual credentials instance.

Signed-off-by: Nicolas De Loof <nicolas.deloof@gmail.com>

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -311,7 +311,7 @@ public class SSHLauncher extends ComputerLauncher {
     public SSHLauncher(String host, int port, String credentialsId,
              String jvmOptions, String javaPath, String prefixStartSlaveCmd, String suffixStartSlaveCmd,
              Integer launchTimeoutSeconds, Integer maxNumRetries, Integer retryWaitTime, SshHostKeyVerificationStrategy sshHostKeyVerificationStrategy) {
-        this(host, port, lookupSystemCredentials(credentialsId), jvmOptions, javaPath, null, prefixStartSlaveCmd,
+        this(host, port, credentialsId, jvmOptions, javaPath, null, prefixStartSlaveCmd,
              suffixStartSlaveCmd, launchTimeoutSeconds, maxNumRetries, retryWaitTime, sshHostKeyVerificationStrategy);
     }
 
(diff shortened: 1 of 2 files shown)


ndeloof:
issue with DockerRule "Failed to build image (100): ff2527313f5c"

ndeloof:
```bash
Reading package lists...
Reading package lists...
Building dependency tree...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 openssh-server : Depends: openssh-client (= 1:7.5p1-10) but 1:7.5p1-10ubuntu0.1 is to be installed
                  Recommends: libpam-systemd but it is not going to be installed
                  Recommends: ncurses-term but it is not going to be installed
                  Recommends: xauth but it is not going to be installed
                  Recommends: ssh-import-id but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
The command '/bin/sh -c apt-get update -y &&     apt-get install -y         openssh-server=1:7.5p1-10' returned a non-zero code: 100
```

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -853,7 +863,7 @@ public Boolean call() throws InterruptedException {
         final Node node = computer.getNode();
         final String nodeName = node != null ? node.getNodeName() : "unknown";
         if(node != null) {
-            CredentialsProvider.track(node, credentials);
+            CredentialsProvider.track(node, getCredentials());
```
oleg-nenashev: Calling a non-final getter in the constructor is generally a bad idea. SSHLauncher is extended in several plugins, and this method is overridden: https://github.com/jenkinsci/docker-plugin/blob/5a2a123cd808a9c82d154cbcc22ef906a6a97907/src/main/java/io/jenkins/docker/connector/DockerComputerSSHConnector.java#L417-L434

oleg-nenashev:
@reviewbybees tho

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
ndeloof: this is not constructor but `launch` method body.
Maybe you've been confused by github UI to hardly capture the context with a Callable involved

PR #86, opened 2018-03-13
developer roles

kuisathaverat:
Are you ok if I set your role as maintainer retired for this plugin?
@stephenc 
@kohsuke 
@olamy 
@mc1arke 
@oleg-nenashev


oleg-nenashev:
fine with me

PR #87, opened 2018-06-12
[WIP][JENKINS-51876] Updated core, Java version, and dependencies

kuisathaverat:
[JENKINS-51876](https://issues.jenkins-ci.org/browse/JENKINS-51876)

The goal is to sanitize the plugin, it supports a really old core and contains tons of methods deprecated since eons, this is the first step to clean up some stuff.

* Update core to 2.60.3
* Update Java to 8
* Update dependencies to latest stable version
* remove trilead-ssh provided dependency, not need because we upgrade the core
* fix TrustHostKeyActionTest Test, I have removed the reflection to create the SSH server is no longer need, this change is moticated to change from a DSA key to an RSA key required for Java 8.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>3.0</version>
+    <version>3.8</version>
   </parent>
 
   <artifactId>ssh-slaves</artifactId>
@@ -24,8 +24,8 @@
   </licenses>
 
   <properties>
-    <jenkins.version>1.625</jenkins.version>
-    <java.level>7</java.level>
+    <jenkins.version>2.60.3</jenkins.version>
+    <java.level>8</java.level>
   </properties>
 
   <developers>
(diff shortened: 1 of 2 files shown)


kuisathaverat:
relaunch CI

pom.xml
```diff
@@ -24,8 +24,8 @@
   </licenses>
 
   <properties>
-    <jenkins.version>1.625</jenkins.version>
-    <java.level>7</java.level>
+    <jenkins.version>2.73.1</jenkins.version>
```
dwnusbaum: Maybe better to use 2.60.3 as the oldest Java 8 LTS unless you need something specific from the 2.73 line.

kuisathaverat: That was my first option, but the thing is that the plugins was compiled for that version by the CI since December, and I found that it is not the same to compile for 2.60.3 or for 2.73.1, I learn it on the hard way in this PR.

https://github.com/jenkinsci/ssh-slaves-plugin/commit/77ed6569ea7e72a87029709bc288efcbe8a8c7a5#diff-58231b16fdee45a03a4ee3cf94a9f2c3

dwnusbaum: FWIW `jenkinsVersions: [null, '2.73.1']` will build the plugin using the baseline in the pom and 2.73.1, so it should have been building on 1.625 and 2.73.1. I'm surprised there would be a big difference between 2.60 and 2.73 though.

PR #88, opened 2018-07-15
[JENKINS-52015] Separate AccessControlled from CredentialsProvider co…

daniel-beck:
…ntext

---

[JENKINS-52015](https://issues.jenkins-ci.org/browse/JENKINS-52015)

Also address the weird side effect that during agent creation, there's no `AccessControlled` other than `Jenkins`, so users who only have Agent/Create would need to create the agent first, and then, in a separate step, configure the SSH launcher.

Needs Jenkins 2.78+ so that `Computer` is a `DescriptorByNameOwner`.

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -1644,7 +1644,8 @@ public class SSHLauncher extends ComputerLauncher {
                                                      @QueryParameter String host,
                                                      @QueryParameter String port,
                                                      @QueryParameter String credentialsId) {
-            if (context == null || !context.hasPermission(Computer.CONFIGURE)) {
+            Jenkins jenkins = Jenkins.getActiveInstance();
+            if ((context == jenkins && !jenkins.hasPermission(Computer.CREATE)) || (context != jenkins && !context.hasPermission(Computer.CONFIGURE))) {
                 return new StandardUsernameListBoxModel()
                         .includeCurrentValue(credentialsId);
             }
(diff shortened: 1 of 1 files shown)


kuisathaverat:
@daniel-beck  I will push the core version to 2.73.1 but I do not mind to push it to 2.78

https://github.com/jenkinsci/ssh-slaves-plugin/pull/87

daniel-beck:
@kuisathaverat No need, unless you want a test for this.

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -1640,12 +1640,12 @@ public String getHelpFile(String fieldName) {
...
 
-        public ListBoxModel doFillCredentialsIdItems(@AncestorInPath ItemGroup context,
+        public ListBoxModel doFillCredentialsIdItems(@AncestorInPath AccessControlled context,
                                                      @QueryParameter String host,
                                                      @QueryParameter String port,
                                                      @QueryParameter String credentialsId) {
-            AccessControlled _context = (context instanceof AccessControlled ? (AccessControlled) context : Jenkins.getInstance());
-            if (_context == null || !_context.hasPermission(Computer.CONFIGURE)) {
+            Jenkins jenkins = Jenkins.getActiveInstance();
+            if ((context == jenkins && !jenkins.hasPermission(Computer.CREATE)) || (context != jenkins && !context.hasPermission(Computer.CONFIGURE))) {
```
daniel-beck: This might result in a regression on Jenkins before 2.78 where it no longer is enough to have system-wide `Computer.CONFIGURE` to get credentials.

PR #90, opened 2018-07-19
[JENKINS-52660] Allow to disable TCP_NODELAY

kuisathaverat:
[JENKINS-52660](https://issues.jenkins-ci.org/browse/JENKINS-52660)

* Configuration option to enable/disable TCP_NODELAY

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -295,9 +295,9 @@ public class SSHLauncher extends ComputerLauncher {
     private final SshHostKeyVerificationStrategy sshHostKeyVerificationStrategy;
 
     /**
-     * Allow to disable the TCP_NODELAY flag on the SSH connection.
+     * Allow to anable/disable the TCP_NODELAY flag on the SSH connection.
      */
-    private boolean disableTcpNoDelay;
+    private Boolean tcpNoDelay;
 
     /**
      * Constructor SSHLauncher creates a new SSHLauncher instance.
@@ -582,6 +582,13 @@ public class SSHLauncher extends ComputerLauncher {
 
     public SSHLauncher(String host, int port, String username, String password, String privatekey, String jvmOptions) {
         this(host,port,username,password,privatekey,jvmOptions,null, null, null);
+    }
+
+    public Object readResolve(){
+        if(tcpNoDelay == null){
+            tcpNoDelay = true;
+        }
+        return this;
     }
 
     public String getCredentialsId() {
@@ -1318,7 +1325,7 @@ public class SSHLauncher extends ComputerLauncher {
     protected void openConnection(final TaskListener listener, final SlaveComputer computer) throws IOException, InterruptedException {
         PrintStream logger = listener.getLogger();
         logger.println(Messages.SSHLauncher_OpeningSSHConnection(getTimestamp(), host + ":" + port));
-        connection.setTCPNoDelay(!disableTcpNoDelay);
+        connection.setTCPNoDelay(getTcpNoDelay());
 
         int maxNumRetries = this.maxNumRetries == null || this.maxNumRetries < 0 ? 0 : this.maxNumRetries;
         for (int i = 0; i <= maxNumRetries; i++) {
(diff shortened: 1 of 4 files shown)


src/main/resources/hudson/plugins/sshslaves/SSHConnector/help-disableTcpNoDelay.html
```diff
@@ -0,0 +1,6 @@
+<div>
+    Disabled the TCP_NODELAY flag for the SSH connection.
```
jvz: Suggested improvements:

> Disables the TCP_NODELAY flag on the SSH connection. If set, data are buffered before being sent through SSH connections. This avoids the frequent sending of small packets which results in poor utilization of the network.

src/main/resources/hudson/plugins/sshslaves/SSHConnector/config.jelly
```diff
@@ -40,5 +40,9 @@
           <f:textbox name="retryWaitTime" />
       </f:entry>
 
+      <f:entry title="${%Disable TCP_NODELAY flag on the SSH connection}" field="disableTcpNoDelay">
+          <f:checkbox name="disableTcpNoDelay" />
+      </f:entry>
```
jeffret-b: Does this retain the prior default? I'm not familiar enough with Jenkins jelly components to be sure how that works.

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -1312,7 +1318,7 @@ protected String checkJavaVersion(final PrintStream logger, String javaCommand,
     protected void openConnection(final TaskListener listener, final SlaveComputer computer) throws IOException, InterruptedException {
         PrintStream logger = listener.getLogger();
         logger.println(Messages.SSHLauncher_OpeningSSHConnection(getTimestamp(), host + ":" + port));
-        connection.setTCPNoDelay(true);
+        connection.setTCPNoDelay(!disableTcpNoDelay);
```
jeffret-b: I find the double negative here kind of confusing.

src/main/resources/hudson/plugins/sshslaves/SSHConnector/config.jelly
kuisathaverat: yes, because the default value of a boolean is false

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
kuisathaverat: agree, every thing sound weird and complex, I'm gonna re-think all again.

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -578,6 +584,13 @@ public SSHLauncher(String host, int port, String username, String password, Stri
         this(host,port,username,password,privatekey,jvmOptions,null, null, null);
     }
 
+    public Object readResolve(){
```
kuisathaverat: it makes old configuration to have the default value TRUE

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -1613,6 +1626,15 @@ public Integer getRetryWaitTime() {
         return retryWaitTime;
     }
 
+    public boolean getTcpNoDelay() {
+        return tcpNoDelay != null ? tcpNoDelay : true;
```
kuisathaverat: it is not possible to have a null value because of the readResolve method, but it is better to protect it

src/main/resources/hudson/plugins/sshslaves/SSHConnector/help-tcpNoDelay.html
```diff
@@ -0,0 +1,7 @@
+<div>
+    Enable/Disables the TCP_NODELAY flag on the SSH connection.
+    If set, disable the Nagle algorithm. This means that segments are always sent as soon as possible,
```
kuisathaverat: it come from https://linux.die.net/man/7/tcp

src/main/resources/hudson/plugins/sshslaves/SSHConnector/help-tcpNoDelay.html
```diff
@@ -0,0 +1,7 @@
...
+    even if there is only a small amount of data. When not set,
+    data is buffered until there is a sufficient amount to send out,
```
jvz: "data" is plural, hence why I said "data are buffered" ;)

dwnusbaum: For any non-native speakers, there is controversy over the correct usage: https://en.wikipedia.org/wiki/Data_(word). I personally consider it an uncountable noun and prefer `data is` :)

kuisathaverat: this text is from the man page, so It is not my fault ;)

kuisathaverat:
relaunch CI

src/main/resources/hudson/plugins/sshslaves/SSHConnector/help-tcpNoDelay.html
jvz: All good! :)

PR #91, opened 2018-07-20
[JENKINS-51876] inform about using a deprecated constructor

kuisathaverat:
[JENKINS-51876](https://issues.jenkins-ci.org/browse/JENKINS-51876)

* Adds a log message on deprecated constructors
* Deprecate constructor that should be deprecated also


kuisathaverat:
my idea, it is to keep them for a couple of version and start to remove the old ones.

PR #92, opened 2018-07-27
[JENKINS-52739] set default values for retries and timeout settings

kuisathaverat:
[JENKINS-52739](https://issues.jenkins-ci.org/browse/JENKINS-52739)

* set DEFAULT_MAX_NUM_RETRIES = 10 tries
* set DEFAULT_RETRY_WAIT_TIME = 15 seconds
* set DEFAULT_LAUNCH_TIMEOUT_SECONDS = 10 (tries) * 15 (retry wait) + 60 seconds
* Fix un treated exceptions types on timeout check


kuisathaverat:
relaunch CI

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -866,7 +881,7 @@ public Boolean call() throws InterruptedException {
             Boolean res;
             try {
                 res = results.get(0).get();
-            } catch (ExecutionException e) {
+            } catch (Exception e) {
```
kuisathaverat: CancelException is not treated, probably more, it breaks the retries logic, by catching the Exception the CancelException is treated.

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -866,8 +881,11 @@ public Boolean call() throws InterruptedException {
             Boolean res;
             try {
                 res = results.get(0).get();
-            } catch (ExecutionException e) {
+            } catch (Throwable e) {
```
jeffret-b: I'm a little concerned about catching Throwable here. It's not generally a good idea. Is this one of those few places where Throwable makes sense? Would it be better to narrow it down to not catch catastrophic failures?

kuisathaverat: yep, I will try to figure out which exceptions could be throw here, the main issue here is, if the exception is not catched, the retries start again from the max number of retries.

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -141,6 +142,9 @@
     private static final List<String> RECOVERABLE_FAILURES = Arrays.asList(
             "Connection refused", "Connection reset", "Connection timed out", "No route to host", "Premature connection close"
     );
+    public static final Integer DEFAULT_MAX_NUM_RETRIES = 10;
```
jvz: Wouldn't it make more sense to use `int` here? The member fields can be nullable, but these constants certainly aren't, and autoboxing will be performed regardless.

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -1415,7 +1434,7 @@ private long getLaunchTimeoutMillis() {
      * @return maxNumRetries
      */
     public Integer getMaxNumRetries() {
```
jvz: Are you stuck with API compatibility here, or can you change these getters to use an `int` instead? Or add `@NonNull` to them at least? :)

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
kuisathaverat: I'm gonna check it because I have changed it due findbugs errors related to boxing-unboxing

kuisathaverat: also, I will need an Integer when I refactor the constructors and use the DataBoundSetter methods

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
kuisathaverat: done

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
kuisathaverat: That's the reason why i set the constants as Integer

At SSHLauncher.java:[line 1439] BX_UNBOXING_IMMEDIATELY_REBOXED

```
    public Integer getMaxNumRetries() {
        return maxNumRetries == null || maxNumRetries < 0 ? DEFAULT_MAX_NUM_RETRIES : maxNumRetries;
    }
```

I can change it to but it is ugly to transform something can be and Integer, I preferred to avoid the transformation.

```
    public Integer getMaxNumRetries() {
        return maxNumRetries == null || maxNumRetries < 0 ? Integer.valueOf(DEFAULT_MAX_NUM_RETRIES) : maxNumRetries;
    }
```

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -881,8 +881,11 @@ public Boolean call() throws InterruptedException {
             Boolean res;
             try {
                 res = results.get(0).get();
-            } catch (Exception e) {
+            } catch (Throwable e) {
                 res = Boolean.FALSE;
+                System.out.println(e.getMessage());
+                listener.getLogger().println("ECHO");
```
olivergondza: Looks like forgotten debugging statements...

kuisathaverat: yep, it is removed on latest commit ;)

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
jvz: Ah, ok, no worries. This would require changing far more signatures clearly; not worth it.

jvz:
You've got a typo in the title, otherwise all good.

PR #97, opened 2018-08-07
JENKINS-37152 Replace 'true' with 'echo'

jcoeltjen:
Use echo > null instead of true command in `verifyNoHeaderJunk()`

Using echo instead of true allows to start windows slaves without cygwin.

This is related to ticket: JENKINS-37152

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -967,7 +967,7 @@ public class SSHLauncher extends ComputerLauncher {
      */
     private void verifyNoHeaderJunk(TaskListener listener) throws IOException, InterruptedException {
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
-        connection.exec("echo hello > null",baos);
+        connection.exec("exit 0",baos);
         final String s;
         //TODO: Seems we need to retrieve the encoding from the connection destination
         try {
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -967,7 +967,7 @@ private EnvVars getEnvVars(DescribableList<NodeProperty<?>, NodePropertyDescript
      */
     private void verifyNoHeaderJunk(TaskListener listener) throws IOException, InterruptedException {
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
-        connection.exec("true",baos);
+        connection.exec("echo hello > null",baos);
```
kuisathaverat: this create a file named null, Is this the intention?

kuisathaverat: maybe it is better to make echo "", it does not create anything and does not print anything on the terminal.

jcoeltjen: Using `echo ''` does not work as it produces output on both linux and windows.
The linux output is only a newline while the windows output would be `''` with a newline.

I will try some other commands I have in mind and will come back to you later.

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -967,7 +967,7 @@ private EnvVars getEnvVars(DescribableList<NodeProperty<?>, NodePropertyDescript
      */
     private void verifyNoHeaderJunk(TaskListener listener) throws IOException, InterruptedException {
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
-        connection.exec("echo hello > null",baos);
+        connection.exec("exit 0",baos);
```
jcoeltjen: I did not try the most obvious choice...
This also works for Windows and Linux agents at the same time and does not leave garbage files on the remote host.

PR #101, opened 2018-09-04
[FIXED JENKINS-53254] Restore getJavaPath() method.

rhencke:
[JENKINS-53254](https://issues.jenkins-ci.org/browse/JENKINS-53254)

It was removed in commit 118423c3de0fd3f5b65183a3f2d62cb2e15a79ca, but
it is still used by at least one plugin.

Mark as deprecated to discourage further use.


oleg-nenashev:
@kuisathaverat Would it be possible to get this fix released?

kuisathaverat:
@oleg-nenashev it will be fixed in a few hours ;), I would release a minor version with the change right now

PR #112, opened 2018-12-10
[JENKINS-54934] Number of retries and retry wait time of 0

offa:
Enables `0`  for the *number of retries* and *retry wait time* (JENKINS-54934).


offa:
@jvz unit tests follow in a few minutes.

If you want to test it manually: Open the configuration of a SSH connected Agent, set those values to `0`, save and reload the page. It's ok, if the `0`'s remain (they were changed to defaults previously).

PR #114, opened 2018-12-28
[JENKINS-55353] Upgrade Jenkins core to 2.150.1

kuisathaverat:
[JENKINS-55353](https://issues.jenkins-ci.org/browse/JENKINS-55353)

After tested the plugin to compile against different Jenkins core versions, I have detected some changes that I have to make to test it.

* Update Jenkins file to compile an test on Jenkins core 2.150.1, 2.164.1, and 2.168 (2 LTS and last weekly)
* Compile and test weekly with JDK 11
* Upgrade Jenkins core to 2.150.1, I will support 2 LTS versions
* Remove JDKInstaler references, is not used in ssh-slaves and it is no longer in Jenkins core, it is in tool-jdk plugin
* Remove deprecated constructors, all these constructors were marked with a warning in logs 6 months ago.
* Remove unused code
* Remove old fields related to deprecated constructors
* Fix tests after changes
* Remove deprecated package dependency javax.xml.bind
* fix Inconsistent synchronization of hudson.plugins.sshslaves.SSHLauncher.host

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+### 1.30.0 (TBD)
+* [JENKINS-55353](https://issues.jenkins-ci.org/browse/JENKINS-55353)
+* [JENKINS-5149](https://issues.jenkins-ci.org/browse/JENKINS-5149) Overwrite remoting.jar only when necessary.
+
+#### Breaking changes
+* this version remove deprecated methods on SSHLauncher see [JENKINS-55353](https://issues.jenkins-ci.org/browse/JENKINS-55353)
+* Support 2.138.4+ LTS Jenkins Core versions
+
 ### 1.29.4 (Dec 28, 2018)
 
 * [JENKINS-54884](https://issues.jenkins-ci.org/browse/JENKINS-54884) [JENKINS-54746](https://issues.jenkins-ci.org/browse/JENKINS-54746) Use Trilead ssh2 from Jenkins Core and use Trilead API plugin for test
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,12 @@
 # Changelog
 
 ### 1.30.0 (TBD)
-* [JENKINS-55353](https://issues.jenkins-ci.org/browse/JENKINS-55353)
+* [JENKINS-55353](https://issues.jenkins-ci.org/browse/JENKINS-55353) Upgrade Jenkins core to 2.150.1
 * [JENKINS-5149](https://issues.jenkins-ci.org/browse/JENKINS-5149) Overwrite remoting.jar only when necessary.
 
 #### Breaking changes
 * this version remove deprecated methods on SSHLauncher see [JENKINS-55353](https://issues.jenkins-ci.org/browse/JENKINS-55353)
-* Support 2.138.4+ LTS Jenkins Core versions
+* Support 2.150.1+ LTS Jenkins Core versions
 
 ### 1.29.4 (Dec 28, 2018)
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,8 +1,8 @@
 buildPlugin(configurations: [
   [ platform: "linux", jdk: "8", jenkins: null ],
   [ platform: "windows", jdk: "8", jenkins: null ],
-  [ platform: "linux", jdk: "8", jenkins: "2.150.1", javaLevel: "8" ],
-  [ platform: "windows", jdk: "8", jenkins: "2.150.1", javaLevel: "8" ],
-  [ platform: "linux", jdk: "11", jenkins: "2.156", javaLevel: "8" ],
-  [ platform: "windows", jdk: "11", jenkins: "2.156", javaLevel: "8" ]
+  [ platform: "linux", jdk: "8", jenkins: "2.150.3", javaLevel: "8" ],
+  [ platform: "windows", jdk: "8", jenkins: "2.150.3", javaLevel: "8" ],
+  [ platform: "linux", jdk: "11", jenkins: "2.164", javaLevel: "8" ],
+  [ platform: "windows", jdk: "11", jenkins: "2.164", javaLevel: "8" ]
 ])
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,8 @@
 buildPlugin(configurations: [
   [ platform: "linux", jdk: "8", jenkins: null ],
-  [ platform: "windows", jdk: "8", jenkins: null ]
+  [ platform: "windows", jdk: "8", jenkins: null ],
+  [ platform: "linux", jdk: "8", jenkins: "2.150.1", javaLevel: "8" ],
+  [ platform: "windows", jdk: "8", jenkins: "2.150.1", javaLevel: "8" ],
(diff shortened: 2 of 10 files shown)


Jenkinsfile
```diff
@@ -1,4 +1,8 @@
 buildPlugin(configurations: [
   [ platform: "linux", jdk: "8", jenkins: null ],
-  [ platform: "windows", jdk: "8", jenkins: null ]
-])
+  [ platform: "windows", jdk: "8", jenkins: null ],
+  [ platform: "linux", jdk: "8", jenkins: "2.150.1", javaLevel: "8" ],
+  [ platform: "windows", jdk: "8", jenkins: "2.150.1", javaLevel: "8" ],
+  [ platform: "linux", jdk: "11", jenkins: "2.156", javaLevel: "8" ],
```
oleg-nenashev: better to use 2.164 which will become lts soon

pom.xml
```diff
@@ -101,6 +101,18 @@
...
+      <exclusions>
+        <exclusion>
+          <groupId>net.i2p.crypto</groupId>
+          <artifactId>eddsa</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>net.i2p.crypto</groupId>
+      <artifactId>eddsa</artifactId>
```
oleg-nenashev: should it be a dependency on the sshd-module instead?

kuisathaverat: yep, it should change in trilead

Jenkinsfile
kuisathaverat: sure

src/main/java/hudson/plugins/sshslaves/SSHConnector.java
```diff
@@ -67,242 +66,187 @@
...
-     *
-     * @deprecated
-     */
-    @Deprecated
-    public transient String privatekey;
-
     /**
      * Field jvmOptions.
      */
-    public final String jvmOptions;
```
KostyaSha: This fields wasn't deprecated and now they are removed

KostyaSha:
> * Upgrade Jenkins core to 2.150.1, I will support 2 LTS versions

This plugin has 226k installations, from long jenkins experience 1 year was almost enough for a lot of companies and users. Huge systems can't do rapid updates to the latest versions. Both 2 LTS full of issues plus plugin broke binary compatibility on MINOR release version.

src/main/java/hudson/plugins/sshslaves/SSHConnector.java
```diff
@@ -67,242 +66,187 @@
...
 
     /**
      *  Field retryWaitTime.
      */
-    public final Integer retryWaitTime;
+    private Integer retryWaitTime;
     
-    private final SshHostKeyVerificationStrategy sshHostKeyVerificationStrategy;
-
-    public StandardUsernameCredentials getCredentials() {
```
KostyaSha: Not deprecated method was removed

src/main/java/hudson/plugins/sshslaves/SSHConnector.java
kuisathaverat: encapsulation, those fields should be accessed by getter/setter methods to avoid mutation

src/main/java/hudson/plugins/sshslaves/SSHConnector.java
kuisathaverat: the constructors that use it was deprecated for months

kuisathaverat:
>This plugin has 226k installations, from long jenkins experience 1 year was almost enough for a lot of companies and users. Huge systems can't do rapid updates to the latest versions. Both 2 LTS full of issues plus plugin broke binary compatibility on MINOR release version.

If they can not upgrade to the latest versions they can not upgrade to the latest version of this plugin, it is a latest version and `Huge systems can't do rapid updates to the latest versions` 

> Both 2 LTS full of issues plus plugin broke binary compatibility on MINOR release version.

You have 7 Jenkins core versions to choose in which one you want to be, those allow you update to the latest version of the plugin.

* 2.150.1 (2018-12-05) 
* 2.150.2 (2019-01-16) 
* 2.150.3 (2019-02-13)
* 2.164.1 (2019-03-13)
* 2.164.2 (2019-04-10)
* 2.164.3 (2019-05-09)
* 2.176.1 (2019-06-10)

KostyaSha:
> You have 7 Jenkins core versions to choose in which one you want to be
.1 and .2 vs .3 known to have bugs, so there are only 2 versions and both.
https://github.com/jenkinsci/jenkins/commits/master/core/src/main/java/jenkins/model/Nodes.java
Different bugs are critical for different systems, Nodes has regressions and fixes, so when i have 2.107.3 i wouldn't update to any of this versions because they are known to be unstable for prod. 

Though core update may not so big issue because people will just not use newer plugin version, but binary compatibility is now broken for **non deprecated** methods.

src/main/java/hudson/plugins/sshslaves/SSHConnector.java
KostyaSha: In jenkins world it's impossible just to remove something existing and public because it will break ALL things that already depending on this. That's why new is done right, old is untouched. Binary compatibility is (was?) one of the main jenkins features that developers/maintainers made a lot of efforts.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+### 1.30.0 (TBD)
+* [JENKINS-55353](https://issues.jenkins-ci.org/browse/JENKINS-55353) Upgrade Jenkins core to 2.150.1
+* [JENKINS-5149](https://issues.jenkins-ci.org/browse/JENKINS-5149) Overwrite remoting.jar only when necessary.
+
+#### Breaking changes
+* this version remove deprecated methods on SSHLauncher see [JENKINS-55353](https://issues.jenkins-ci.org/browse/JENKINS-55353)
```
daniel-beck, line 8: The issue does not explain why this was done at all.

kuisathaverat: This was a long time ago, about 4 (or more) years, I deprecated a bunch of constructors, after a time enough to make pluggins the change, IIRC a year or so, those deprecated constructors were removed. The reason to remove those constructors is because there were many and it was a mess to have to maintain 5-10 constructors deprecated every time you change something in a filed, also maintain test to check that those Contructor s still works, so for maintenance reasons I have decided to remove them.

kuisathaverat: Also, constructors with ten or more parameters where only three are mandatory does not have any sense

daniel-beck: I'm not arguing there's no possible reason for the change, I'm pointing out the supposed justification is missing. At least that's my reading of a changelog entry that points at a Jira issue.

And the explanation about warnings only help with production code (which BTW is generally not something I'd recommend, warnings are seen only by admins, potentially extremely frequently, while not something they can do anything about other than report bugs). Modernizing `groovy` plugin and adapting its tests can still happen years later.

kuisathaverat: Yep, maybe a warning in the logs was not the best way to a ounce the change, but it was in the release notes too, and when the change become a breaking change it was in the release notes as breaking change, the change was four years ago, I do not remember the details but IIRC I search for plugins using those constructors and made some PRs, but I really donor remember it, it was four year ago

PR #116, opened 2019-01-30
[FIX JENKINS-56147] Overwrite remoting.jar only when necessary.

lgrill-pentaho:
When an ssh agent is launched it will compare the MD5 hash of the remoting jar in Jenkins home with the hash of the remoting.jar in the remote root directory of the agent.

* If the file does not exist or is different, it is copied from the Jenkins master.
* If the file is the same, it is ignored.

This change allows multiple agent nodes to share the same root directory (and workspace) without getting concurrency conflicts that happened when the remoting jar got overwritten with every agent launch.

See https://issues.jenkins-ci.org/browse/JENKINS-5149 and similar.

see the related issue [JENKINS-56147](https://issues.jenkins-ci.org/browse/JENKINS-56147)

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -1080,7 +1080,7 @@ public class SSHLauncher extends ComputerLauncher {
                 SFTPv3FileAttributes fileAttributes = sftpClient._stat(workingDirectory);
                 if (fileAttributes==null) {
                     listener.getLogger().println(Messages.SSHLauncher_RemoteFSDoesNotExist(getTimestamp(),
-                      workingDirectory));
+                            workingDirectory));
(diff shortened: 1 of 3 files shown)


lgrill-pentaho:
Sorry, but this change does fix the issue and is well proven in a production environment on-site. We are aware of the potential concurrency problems when using the same path in a shared workspace, so with the exception of a shared local maven cache, our builds never do that.

kuisathaverat:
if you are aware that this does not resolve the concurrence issue, it only copy the remoting file when it changes, I would accept the PR but It needs a Jira issue open as improvement describing the change, to have a trace of the change.

**this change NOT GUARANTEE that you can run several agents in the same host sharing the working directory, this behavior is NOT SUPPORTED and NOT RECOMMENDED.**

kuisathaverat:
@lgrill-pentaho feel free to assign to yourself the Jira issue, I did not find your user.

PR #117, opened 2019-02-04
[JENKINS-55787] Switch from entry to optional block

jsoref:


jsoref:
Please hold off on this, I need to work on the migration bits :-(.

jsoref:
https://github.com/jenkinsci/ssh-plugin/blob/f176422f9284d77da1cd745ae2c8a9119cc97b8a/src/main/resources/org/jvnet/hudson/plugins/SSHBuildWrapper/global.jelly#L19-L20

Here's an example of how this looks today:
image

jsoref:
Here's an example of the current element:
image

jsoref:
The pictures below are using the latest version of this PR (which doesn't use optional-block):

here's with the current Jenkins:
image

Of course, the more frustrating piece is that the help layout is such that I'm pretty sure the reason people picked the style I'm moving away from is that the help widget is wider than the content.

w/ my alternate UI engine:
image

Sadly, at this time the help widget can't be shown by relying on an attribute directly on the f:checkbox.

It looks like I'll probably try to land the core side first, let things look ugly for a bit and then arrange for people to merge the corresponding cleanup.

Note that standard UX convention has checkboxes to the left of labels (for LTR languages), so if people are comfortable landing this change in advance of core change, it is generally an improvement (the help thing is a minor botch, and not really terrible).

-- Config changing works.

PR #123, opened 2019-04-08
[JENKINS-57018] Allow to enable/disable TCP_NODELAY in SSHConnector

thaiphv:
This utilizes the change introduced by #90. Essentially it allows the new setting can be set for instances of SSHConnector. Effectively, we can change and persist this setting in the ec2-fleet-plugin when SSHConnector is used as a launcher.

[JENKINS-57018](https://issues.jenkins-ci.org/browse/JENKINS-57018)

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/SSHConnector.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHConnector.java
@@ -107,8 +107,13 @@ public class SSHConnector extends ComputerConnector {
      *  Field retryWaitTime.
      */
     private Integer retryWaitTime;
-    
+
     private SshHostKeyVerificationStrategy sshHostKeyVerificationStrategy;
+
+    /**
+     *  Field tcpNoDelay.
+     */
+    private Boolean tcpNoDelay;
 
     /**
      * Constructor SSHLauncher creates a new SSHLauncher instance.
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/sshslaves/SSHConnector.java
```diff
@@ -107,9 +107,14 @@
...
      */
     private Integer retryWaitTime;
-    
+
     private SshHostKeyVerificationStrategy sshHostKeyVerificationStrategy;
 
+    /**
+     *  Field tcpNoDelay.
+     */
+    private Boolean tcpNoDelay;
```
kuisathaverat:
```suggestion
    private boolean tcpNoDelay = true;
```

src/main/java/hudson/plugins/sshslaves/SSHConnector.java
```diff
@@ -247,6 +259,10 @@ public Integer getRetryWaitTime() {
         return retryWaitTime;
     }
 
+    public Boolean getTcpNoDelay() {
```
kuisathaverat: if we use boolean primitive with default value we do not need this logic, but we can keep the getter

```suggestion
    public boolean getTcpNoDelay() {
```

src/main/java/hudson/plugins/sshslaves/SSHConnector.java
```diff
@@ -247,6 +259,10 @@ public Integer getRetryWaitTime() {
...
+        return tcpNoDelay != null ? tcpNoDelay : true;
```
kuisathaverat:
```suggestion
        return tcpNoDelay;
```

src/main/java/hudson/plugins/sshslaves/SSHConnector.java
```diff
@@ -202,7 +209,12 @@ public void setSshHostKeyVerificationStrategy(SshHostKeyVerificationStrategy val
     public void setPort(int value){
         this.port = value == 0 ? DEFAULT_SSH_PORT : value;
     }
-    
+
+    @DataBoundSetter
+    public void setTcpNoDelay(Boolean tcpNoDelay) {
```
kuisathaverat:
```suggestion
    public void setTcpNoDelay(boolean tcpNoDelay) {
```

src/main/java/hudson/plugins/sshslaves/SSHConnector.java
```diff
@@ -202,7 +209,12 @@ public void setSshHostKeyVerificationStrategy(SshHostKeyVerificationStrategy val
...
+        this.tcpNoDelay = tcpNoDelay;
```
kuisathaverat: but we need it here :palmface

src/main/java/hudson/plugins/sshslaves/SSHConnector.java
kuisathaverat: forget it

src/main/java/hudson/plugins/sshslaves/SSHConnector.java
kuisathaverat: forget it

src/main/java/hudson/plugins/sshslaves/SSHConnector.java
kuisathaverat: forget it

kuisathaverat:
I'll create an Jira issue to track it and I will merge it

thaiphv:
@kuisathaverat, thanks.

PR #126, opened 2019-05-31
[JENKINS-57797] Support Configuration-as-Code

alecharp:
[JENKINS-57797](https://issues.jenkins-ci.org/browse/JENKINS-57797)

This PR aims to validate the support of configuration-as-code for this plugin.

By default, the configuration is supported, but under the section `sSHLauncher`, this is why I added the `@Symbol` on the `SSHLauncher` descriptor.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -27,6 +27,7 @@
   <properties>
     <jenkins.version>2.150.1</jenkins.version>
     <java.level>8</java.level>
+    <configuration-as-code.version>1.17</configuration-as-code.version>
   </properties>
 
   <developers>
@@ -123,6 +124,19 @@
           <artifactId>commons-io</artifactId>
         </exclusion>
       </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>io.jenkins.configuration-as-code</groupId>
+      <artifactId>configuration-as-code-support</artifactId>
+      <version>1.17</version>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>io.jenkins</groupId>
+      <artifactId>configuration-as-code</artifactId>
+      <version>${configuration-as-code.version}</version>
+      <scope>test</scope>
+      <type>test-jar</type>
     </dependency>
   </dependencies>
   <build>
--- a/pom.xml
+++ b/pom.xml
@@ -128,7 +128,7 @@
     <dependency>
       <groupId>io.jenkins.configuration-as-code</groupId>
       <artifactId>configuration-as-code-support</artifactId>
-      <version>1.17</version>
+      <version>${configuration-as-code.version}</version>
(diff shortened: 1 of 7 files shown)


src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -1223,6 +1224,7 @@ public String getWorkDirParam(@NonNull String workingDirectory){
     }
 
     @Extension
+    @Symbol({"sSHLauncher", "ssh"})
```
oleg-nenashev: With such ordering `sSHLauncher` will be a default alias for export. I suggest keeping it only for compatibility purposes

```suggestion
    @Symbol({"ssh", "sSHLauncher"})
```

alecharp:
it's missing the test about the round-trip export/import of the configuration.

src/test/java/hudson/plugins/sshslaves/SSHLauncherCasCRoundTripTest.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @Override
+    protected String stringInLogExpected() {
+        return "hudson.slaves.DumbSlave.name = this-ssh-agent";
+    }
+
+    @Override
+    protected String configResource() {
+        return "SSHCasCConfig.yml";
```
MRamonLeon: FTR: If you rename the yml file to _configuration-as-code.yaml_ you don't need to override that.

alecharp: I have two files. I prefer each on to have a explicit name.

batmat:
@kuisathaverat I think this is ready for a maintainer review :). Could you please have a look, and possibly merge & release this? 

Gracias!

kuisathaverat:
@batmat @alecharp I'll try to release it this weekend

MRamonLeon:
:+1: We still love you @kuisathaverat

pom.xml
```diff
@@ -27,6 +27,8 @@
   <properties>
     <jenkins.version>2.150.1</jenkins.version>
     <java.level>8</java.level>
+    <!-- TODO fix once released -->
+    <configuration-as-code.version>1.20-20190604.105601-2</configuration-as-code.version>
```
batmat: @kuisathaverat wooops, there's actually an incrementals in use here. I don't think we should release in this state.

kuisathaverat: it is a test dependency, I guess if I try with the latest version and the test pass is ok to release changing the dependency to the latest version.

MRamonLeon: Waiting for the release today, most likely. CC: @oleg-nenashev @casz

PR #130, opened 2019-06-02
update jenkins.version to LTS from 2.150.1 to 2.164.3

based2:
https://jenkins.io/changelog-stable/


kuisathaverat:
I support 2 major LTS versions so this change does not make sense yet

PR #134, opened 2019-06-24
[JENKINS-57126] Support alternative drive letters for Windows ssh agents using PowerShell/Win32-OpenSSH

rkivisto:
As per https://issues.jenkins-ci.org/browse/JENKINS-57126

## Issue

If you are using [PowerShell/Win32-OpenSSH](https://github.com/PowerShell/Win32-OpenSSH/releases/tag/v8.0.0.0p1-Beta) and you set the "Remote root directory" to a drive other than the "C:\" (such as somewhere on the `D:\`), the agent will fail to launch with the error:

```
[04/20/19 09:21:29] [SSH] Starting agent process: cd "D:\jenkins-agent" && "C:\Program Files (x86)\Java\jre1.8.0_201\bin\java.exe" -jar remoting.jar -workDir D:\jenkins-agent
Error: Unable to access jarfile remoting.jar
Slave JVM has terminated. Exit code=1
[04/20/19 09:21:29] Launch failed - cleaning up connection
```
## Root cause

The root cause appears to be that the command `cd "D:\jenkins-agent"` does not successfully change to the `D:\` drive, because when using Windows command prompt, you need to add the `/d` argument `cd` to change the drive:

> /d : Changes the current drive or the current directory for a drive.

As per https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/cd

## Proposed changes

With the changes from this PR, the `/d` argument is passed if the "Remote root directory" appears to be a Windows-like path, and Cygwin is not detected.

## Testing done

I've tested this PR using a fresh instance of Jenkins LTS 2.176.1
I installed the "recommended" plugins from the initial setup wizard, then I installed my snapshot of the ssh-slaves-plugin.

### PowerShell/Win32-OpenSSH Testing
I tested that `PowerShell/Win32-OpenSSH` now supports the `remote root directory` to be on the `D:\` drive (specifically I tested `D:\j`) on a fresh EC2 instance running:
>  Microsoft Windows Server 2019 Base - ami-09ef280df1a6a5330

I installed PowerShell/Win32-OpenSSH version:
>  https://github.com/PowerShell/Win32-OpenSSH/releases/tag/v8.0.0.0p1-Beta

and I setup ssh by following the `On earlier versions of Windows` and also `Configuring SSH server` from:
>  https://winscp.net/eng/docs/guide_windows_openssh_server

The agent connected successfully and the startup logs were:

```
[06/24/19 16:43:40] [SSH] Starting agent process: cd /d "D:\j" && "C:\Program Files (x86)\Java\jre1.8.0_211\bin\java.exe"  -jar remoting.jar -workDir D:\j
Jun 24, 2019 8:43:40 PM org.jenkinsci.remoting.engine.WorkDirManager initializeWorkDir
INFO: Using D:\j\remoting as a remoting work directory
Both error and output logs will be printed to D:\j\remoting

<===[JENKINS REMOTING CAPACITY]===>channel started
Remoting version: 3.29
This is a Windows agent

Agent successfully connected and online
```

### Cygwin testing

I also tested that Cygwin ssh agents are not affected by this change by starting a fresh EC2 instance running:
>  Microsoft Windows Server 2019 Base - ami-09ef280df1a6a5330

and I setup ssh by following:
>  https://go.cloudbees.com/docs/cloudbees-core/cloud-admin-guide/agents/#_microsoft_windows

then I confirmed that the "/d" argument is not passed to "cd" for Cygwin ssh agents.

### Testing to disable this feature

I also tested that if I start Jenkins with:

```
 java -Dhudson.plugins.sshslaves.alternativewindowsdrive=false -jar jenkins.war
```

this feature is successfully disabled.

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -595,14 +595,28 @@ public class SSHLauncher extends ComputerLauncher {
         session = connection.openSession();
         expandChannelBufferSize(session,listener);
 
-        String cdArguments = "";
+        PrintStream logger = listener.getLogger();
+
+         // Code to support alternate drives (such as D:\) for the remote root directory on Windows, when not using cygwin.
+         // You can disable this feature with startup option: -Dhudson.plugins.sshslaves.alternativewindowsdrive=false
+        boolean AlternativeWindowsDriveCheck = ! "false".equals(System.getProperty("hudson.plugins.sshslaves.alternativewindowsdrive"));
+        String agentOSTYPE="";
+        try {
+            ByteArrayOutputStream agentOsTypeStream = new ByteArrayOutputStream();
+            connection.exec("echo \"$OSTYPE\"", agentOsTypeStream);
+            agentOSTYPE = StringUtils.chomp( agentOsTypeStream.toString(Charset.defaultCharset().name()) );
+        } catch (InterruptedException ex) {
+            // exceptions here will only disable the AlternativeWindowsDriveCheck
+            logger.println(Messages.SSHLauncher_UnableToGetEnvironment(getTimestamp()));
+            AlternativeWindowsDriveCheck=false;
+        }
         Pattern windowsDrivePattern = Pattern.compile("^[a-zA-Z]:\\\\");
         Matcher windowsDriveMatcher = windowsDrivePattern.matcher(workingDirectory);
-        if ( windowsDriveMatcher.find() && ! "false".equals(System.getProperty("hudson.plugins.sshslaves.alternativewindowsdrive")) ) {
-            // To support alternate drives for the remote root directory on Windows, when not using cygwin.
-            // You can disable this feature with startup option: -Dhudson.plugins.sshslaves.alternativewindowsdrive=false
+        String cdArguments = "";
+        if ( AlternativeWindowsDriveCheck && ! "cygwin".equals(agentOSTYPE) && windowsDriveMatcher.find() ) {
             cdArguments = "/d ";
         }
+
         String cmd = "cd " + cdArguments + "\"" + workingDirectory + "\" && " + java + " " + getJvmOptions() + " -jar " + AGENT_JAR +
                      getWorkDirParam(workingDirectory);
 
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -234,6 +234,11 @@ public class SSHLauncher extends ComputerLauncher {
     private String workDir;
 
     /**
+     * [JENKINS-57126] Support alternative drive letters for Windows ssh agents using PowerShell/Win32-OpenSSH
+     */
+    private static final Pattern windowsDrivePattern = Pattern.compile("^[a-zA-Z]:\\\\");
+
+    /**
      * Constructor SSHLauncher creates a new SSHLauncher instance.
      *
      * @param host       The host to connect to.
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -592,7 +594,30 @@ private void startAgent(SlaveComputer computer, final TaskListener listener, Str
...
+
+        PrintStream logger = listener.getLogger();
+
+         // Code to support alternate drives (such as D:\) for the remote root directory on Windows, when not using cygwin.
+         // You can disable this feature with startup option: -Dhudson.plugins.sshslaves.alternativewindowsdrive=false
+        boolean AlternativeWindowsDriveCheck = ! "false".equals(System.getProperty("hudson.plugins.sshslaves.alternativewindowsdrive"));
+        String agentOSTYPE="";
+        try {
+            ByteArrayOutputStream agentOsTypeStream = new ByteArrayOutputStream();
+            connection.exec("echo \"$OSTYPE\"", agentOsTypeStream);
```
jvz: Is this value going to differ from the system property value for `os.arch` or the others? That could be simpler than checking the shell like that, though this approach would still work.

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -592,7 +594,30 @@ private void startAgent(SlaveComputer computer, final TaskListener listener, Str
                             String workingDirectory) throws IOException {
         session = connection.openSession();
         expandChannelBufferSize(session,listener);
-        String cmd = "cd \"" + workingDirectory + "\" && " + java + " " + getJvmOptions() + " -jar " + AGENT_JAR +
+
+        PrintStream logger = listener.getLogger();
+
+         // Code to support alternate drives (such as D:\) for the remote root directory on Windows, when not using cygwin.
+         // You can disable this feature with startup option: -Dhudson.plugins.sshslaves.alternativewindowsdrive=false
+        boolean AlternativeWindowsDriveCheck = ! "false".equals(System.getProperty("hudson.plugins.sshslaves.alternativewindowsdrive"));
```
jvz: Can you rename this to start with a lowercase letter?

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -592,7 +594,30 @@ private void startAgent(SlaveComputer computer, final TaskListener listener, Str
...
+        try {
+            ByteArrayOutputStream agentOsTypeStream = new ByteArrayOutputStream();
+            connection.exec("echo \"$OSTYPE\"", agentOsTypeStream);
+            agentOSTYPE = StringUtils.chomp( agentOsTypeStream.toString(Charset.defaultCharset().name()) );
+        } catch (InterruptedException ex) {
+            // exceptions here will only disable the AlternativeWindowsDriveCheck
+            logger.println(Messages.SSHLauncher_UnableToGetEnvironment(getTimestamp()));
+            AlternativeWindowsDriveCheck=false;
+        }
+        Pattern windowsDrivePattern = Pattern.compile("^[a-zA-Z]:\\\\");
```
jvz: You can move this to a `private static final` field to cache the regex.

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -592,7 +594,30 @@ private void startAgent(SlaveComputer computer, final TaskListener listener, Str
...
+        Matcher windowsDriveMatcher = windowsDrivePattern.matcher(workingDirectory);
+        String cdArguments = "";
+        if ( AlternativeWindowsDriveCheck && ! "cygwin".equals(agentOSTYPE) && windowsDriveMatcher.find() ) {
+            cdArguments = "/d ";
+        }
+
+        String cmd = "cd " + cdArguments + "\"" + workingDirectory + "\" && " + java + " " + getJvmOptions() + " -jar " + AGENT_JAR +
```
jvz: Now I know you didn't cause this, but it would be nice to ensure that arguments are properly quoted when constructing raw commands like this. However, since this script seems to be agnostic of bash or batch, I'm not exactly sure what would be the best way to ensure that happens here.

For comparison, the easiest way to ensure user input is quoted properly in a command is to rely on the underlying environment to do that for you. In that case, it'd be to set environment variables with the user input, then executing a shell script that refers to properly quoted forms of the environment variables. That way the shell handles the escaping for you.

For some code you can look into, see https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/jenkins/util/WindowsUtil.java for the Windows side of things at least.

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
rkivisto: Is `os.arch` for the agent available when the remoting connection is just established (at this point of the code)? I had first tried checking properties, but I had found that they were either for the Jenkins Master (not helpful), or they were not initialized for the Jenkins agent yet, hence why I went this route.

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
rkivisto: I think that is a good idea to address in a separate PR

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
rkivisto: @jvz I've addressed all the other comments except this one

src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
```diff
@@ -592,7 +599,29 @@ private void startAgent(SlaveComputer computer, final TaskListener listener, Str
...
+        boolean alternativeWindowsDriveCheck = ! "false".equals(System.getProperty("hudson.plugins.sshslaves.alternativewindowsdrive"));
+        String agentOSTYPE="";
+        try {
+            ByteArrayOutputStream agentOsTypeStream = new ByteArrayOutputStream();
+            connection.exec("echo \"$OSTYPE\"", agentOsTypeStream);
+            agentOSTYPE = StringUtils.chomp( agentOsTypeStream.toString(Charset.defaultCharset().name()) );
+        } catch (InterruptedException ex) {
+            // exceptions here will only disable the alternativeWindowsDriveCheck
+            logger.println(Messages.SSHLauncher_UnableToGetEnvironment(getTimestamp()));
+            alternativeWindowsDriveCheck=false;
```
jvz:
```suggestion
            alternativeWindowsDriveCheck=false;
            Thread.currentThread().interrupt();
```

Catching an interrupt and not resetting the interrupt flag will lead to zombie threads or some other bad behavior, I forget which.

kuisathaverat:
using `cd /d D:\ &&` as suffix works, this thing will be aborded on https://issues.jenkins-ci.org/browse/JENKINS-55362, thank you for your contribution but IMHO it does not worth the additional code, for something is resolved in an easy way. We can add the tip to the Troubleshooting

PR #139, opened 2019-07-21
[JENKINS-56187] manually provided key verification strategy crashes

kuisathaverat:
see [JENKINS-56187](https://issues.jenkins-ci.org/browse/JENKINS-56187)

* catch IllegalArgumentException to avoid show the exception in the UI
* sent the public key using POST method to prevent too long GET request errors

### DIFF
--- a/src/main/resources/hudson/plugins/sshslaves/verifiers/ManuallyProvidedKeyVerificationStrategy/config.jelly
+++ b/src/main/resources/hudson/plugins/sshslaves/verifiers/ManuallyProvidedKeyVerificationStrategy/config.jelly
@@ -35,7 +35,7 @@ THE SOFTWARE.
     </j:if>           
 
 	<f:entry title="${%SSH Key}" field="key">
-		<f:textarea />
+		<f:textarea checkMethod="post"/>
 	</f:entry>
 
 </j:jelly>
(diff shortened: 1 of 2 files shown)


src/main/resources/hudson/plugins/sshslaves/verifiers/ManuallyProvidedKeyVerificationStrategy/config.jelly
```diff
@@ -35,7 +35,7 @@ THE SOFTWARE.
     </j:if>           
 
 	<f:entry title="${%SSH Key}" field="key">
-		<f:textarea />
+		<f:textarea checkMethod="post"/>
```
jvz: This isn't leaking information about the key, is it? I don't see any permission checks.

kuisathaverat: It is a public key so it would not be a security issue, however, I'd open an issue to check every do* to use POST and require proper permission to avoid abuse and security issues.

kuisathaverat: https://issues.jenkins-ci.org/browse/JENKINS-58691

PR #144, opened 2019-07-29
Enable Release Drafter for the repository

oleg-nenashev:
Just a quality of life improvement which simplifies changelog management in the plugin.


oleg-nenashev:
Should start working within few hours after the merge

PR #145, opened 2019-07-29
Move Changelog to GitHub Releases

oleg-nenashev:


oleg-nenashev:
I will need to manually rebuild the changelog for commits before this change (tag resolution glitches), but it should work just fine for all further releases

PR #168, opened 2019-10-13
[JENKINS-47510] RemoteLauncher.isUnix() is not overridden - always returns master OS

kuisathaverat:
See [JENKINS-47510](https://issues.jenkins-ci.org/browse/JENKINS-47510).

### Submitter checklist

- [X] JIRA issue is well described

@jtnord

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/RemoteLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/RemoteLauncher.java
@@ -38,6 +38,7 @@ import hudson.remoting.ChannelBuilder;
 import hudson.util.StreamCopyThread;
 
 import java.io.BufferedOutputStream;
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/sshslaves/RemoteLauncher.java
```diff
@@ -140,4 +141,9 @@ private String makeCommandLine(List<String> cmd, FilePath _workDir) {
     public void kill(Map<String, String> modelEnvVars) throws IOException, InterruptedException {
         // no way to do this
     }
+
+  @Override
+  public boolean isUnix() {
+    return File.pathSeparatorChar == ':';
```
jvz: Is this better than checking `!Functions.isWindows()`?

kuisathaverat: I did not know about that function, it is better to use it.

src/main/java/hudson/plugins/sshslaves/RemoteLauncher.java
```diff
@@ -140,4 +141,9 @@ private String makeCommandLine(List<String> cmd, FilePath _workDir) {
     public void kill(Map<String, String> modelEnvVars) throws IOException, InterruptedException {
         // no way to do this
     }
+
+  @Override
+  public boolean isUnix() {
+    return !Functions.isWindows();
```
jtnord: does this actually run on the agent?
if not this returns the same s of the master which is substituting one big for a regression

kuisathaverat: I've remembered why it was not merged.

PR #170, opened 2020-01-01
Revert "[JENKINS-47510] RemoteLauncher.isUnix() is not overridden - always returns master OS"

kuisathaverat:
Reverts jenkinsci/ssh-slaves-plugin#168


oleg-nenashev:
Why was it removed? Not released so far tho

kuisathaverat:
the code runs on the Master so the OS detection is not good :(

PR #96, opened 2018-08-05
[JENKINS-42840] Slave to Agent Renaming

kuisathaverat:
[JENKINS-42840](https://issues.jenkins-ci.org/browse/JENKINS-42840)

* Change slave.jar to remoting.jar
* Change javadoc
* Change some methods name (startSlave, copySlaveJar)
* Change messages

Pending to change SSH Slaves Plugin to something else, these are the proposals:

* SSH Nodes Plugins
* SSH Agent Nodes Plugin

PR #171, opened 2020-01-02
[JENKINS-42840] - Cleanup the "slave" mention leftovers in the documentation

oleg-nenashev:
See [JENKINS-42840](https://issues.jenkins-ci.org/browse/JENKINS-42840). There were some leftovers after https://github.com/jenkinsci/ssh-slaves-plugin/pull/96

This PR cleans up documentation and Javadoc + some tests while I was around

### Submitter checklist

- [x] JIRA issue is well described
- [x] Appropriate autotests or explanation to why this change has no tests

### DIFF
--- a/doc/CONFIGURE.md
+++ b/doc/CONFIGURE.md
@@ -1,4 +1,5 @@
-# SSH Slaves plugin
+# Configuring the SSH Build Agents plugin
+
 This plugin allows you to manage an agent running on \*nix machines over SSH. It adds a new type of agent launch method.
 This launch method will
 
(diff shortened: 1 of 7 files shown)


doc/CONFIGURE.md
```diff
@@ -200,12 +201,12 @@ and the specified hostname and port, so where you have created the appropriate c
 will be restricted to those outside of any credential domain and those from matching credential domains.
 This can help differentiate between multiple keys/password associated with the same username.
 
-## Using Cygwin
-See [SSH slaves and Cygwin](https://wiki.jenkins.io/display/JENKINS/SSH+slaves+and+Cygwin) for the discussion of how to use this plugin to talk to Cygwin SSHD server.
+### Using Cygwin
```
oleg-nenashev: @kuisathaverat I am thinking about moving this Wiki to plugin docs . Do we really want to do it? Or would you rther prefer to deprecate it in favor of native SSH support by modern Windows versions?

kuisathaverat: I  guess that old windows versions still do not have native SSH support, so this documentation is still valid, thus we should move it to plugin documentation.

oleg-nenashev: OK, will do it in a follow-up pull request

PR #174, opened 2020-01-19
[JENKINS-60812] Failed known_hosts verification for non-standard ssh port

kuisathaverat:
See [JENKINS-60812](https://issues.jenkins-ci.org/browse/JENKINS-60812).

### Submitter checklist

- [X] JIRA issue is well described
- [X] Appropriate autotests or explanation to why this change has no tests

test environment https://github.com/kuisathaverat/jenkins-issues/tree/master/JENKINS-60812

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/verifiers/KnownHostsFileKeyVerificationStrategy.java
+++ b/src/main/java/hudson/plugins/sshslaves/verifiers/KnownHostsFileKeyVerificationStrategy.java
@@ -23,24 +23,21 @@
  */
 package hudson.plugins.sshslaves.verifiers;
 
-import java.io.File;
-import java.io.IOException;
-import java.nio.file.Path;
-import java.nio.file.Paths;
-
+import com.trilead.ssh2.KnownHosts;
+import hudson.Extension;
+import hudson.model.TaskListener;
+import hudson.plugins.sshslaves.Messages;
+import hudson.plugins.sshslaves.SSHLauncher;
 import hudson.slaves.ComputerLauncher;
+import hudson.slaves.SlaveComputer;
 import org.apache.commons.lang.StringUtils;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.DataBoundConstructor;
 
-import com.trilead.ssh2.KnownHosts;
-
-import hudson.Extension;
-import hudson.model.TaskListener;
-import hudson.plugins.sshslaves.Messages;
-import hudson.plugins.sshslaves.SSHLauncher;
-import hudson.slaves.SlaveComputer;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Paths;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/plugins/sshslaves/verifiers/KnownHostsFileKeyVerificationStrategy.java
```diff
@@ -55,64 +52,76 @@
...
 
-        KnownHosts knownHosts = new KnownHosts(KNOWN_HOSTS_FILE);
-        return knownHosts.getPreferredServerHostkeyAlgorithmOrder(((SSHLauncher) launcher).getHost());
+    listener.getLogger().println(Messages.KnownHostsFileHostKeyVerifier_SearchingFor(hostPort, KNOWN_HOSTS_FILE));
+    int resultHostPort = verify(hostPort, hostKey.getAlgorithm(), hostKey.getKey());
+
+    if (KnownHosts.HOSTKEY_IS_OK == resultHost || KnownHosts.HOSTKEY_IS_OK == resultHostPort) {
+      listener.getLogger().println(Messages.KnownHostsFileHostKeyVerifier_KeyTrusted(SSHLauncher.getTimestamp()));
+      return true;
+    } else if (KnownHosts.HOSTKEY_IS_NEW == resultHost && KnownHosts.HOSTKEY_IS_NEW == resultHostPort) {
```
varyvol: Is this and & or an | ? What happens if only one of them is new?

kuisathaverat: if none of both is `HOSTKEY_IS_OK`, it would mean both are HOSTKEY_IS_NEW (so hosts keys are not in known_hosts), or one of them returns an unknown error, so it will report the worst error that happens.

varyvol: More than right, thanks for the clarification!

daniel-beck:
I am not confident that this is a reasonable change. There's got to be a reason that known_hosts limits by port number, and I see no research into why that is, and why it wouldn't matter for Jenkins.

The SSH server in the example was even running on a non-restricted port (>1024); indicating it could have been set up by any user with a shell on that machine.

kuisathaverat:
@daniel-beck if you add a key for a host that uses the `port 22`(ssh user@host), the host will be added with the name of `the host` to `known_hosts`, if you add a host that not uses the port 22, for example, `port 2222`(ssh user@host -p 2222), the host is added as `host:port`, so in the current implementation, the key will not be found because it will search only for the `host`. The current PR, make a change to search for `the host` (current behavior) and the `host:port` to allow find the hosts when the key was for a non-standard port.
A host can have several ports exposed and those can have several different keys, because of that the `known_hosts` file store the host and the key for no 22 ports, it is documented as it in the OpenSSH documentation https://man.openbsd.org/sshd#SSH_KNOWN_HOSTS_FILE_FORMAT

PR #177, opened 2020-02-25
Demonstrate adding findsecbugs to spotbugs.

jeffret-b:
Demonstrate adding findsecbugs to spotbugs. This commit is a demonstration and is not intended to be merged as-is.

Findsecbugs reports several findings for this plugin, but all of them can be safely suppressed or reworked. There are a few that could be a concern, but examination shows that as currently used they do not introduce a vulnerability. The MD5 hash is not stored so it can be easily converted to a more reliable algorithm.

This also uses the exclude file for a couple of finding types that are fairly common in Jenkins but not considered a concern.

---

My plan is to add findsecbugs at the plugin pom after sufficient testing and communication. At that point, it would make sense to add the suppressions, but not the pom file changes from this PR.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -121,7 +121,7 @@
       </dependency>
     </dependencies>
   </dependencyManagement>
-  
+
   <dependencies>
     <!-- plugin dependencies -->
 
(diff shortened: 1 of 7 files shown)


kuisathaverat:
@jeffret-b is this still valid

jeffret-b:
We don't need to leave it open. I'll close it.

PR #178, opened 2020-04-08
cleanupConnection should not block on Connection.close

jglick:
Not the same as #75 or #78.

```
"Computer.threadPoolForRemoting [#…]" … waiting for monitor entry …
   java.lang.Thread.State: BLOCKED (on object monitor)
	at hudson.plugins.sshslaves.SSHLauncher.launch(SSHLauncher.java:402)
	- waiting to lock <0x…> (a hudson.plugins.sshslaves.SSHLauncher)
	at hudson.slaves.SlaveComputer$1.call(SlaveComputer.java:294)
	at …
"Computer.threadPoolForRemoting [#…]" … waiting for monitor entry …
   java.lang.Thread.State: BLOCKED (on object monitor)
	at com.trilead.ssh2.Connection.close(Connection.java:529)
	- waiting to lock <0x…> (a com.trilead.ssh2.Connection)
	at hudson.plugins.sshslaves.SSHLauncher.cleanupConnection(SSHLauncher.java:511)
	at hudson.plugins.sshslaves.SSHLauncher.tearDownConnectionImpl(SSHLauncher.java:1006)
	at hudson.plugins.sshslaves.SSHLauncher.tearDownConnection(SSHLauncher.java:930)
	- locked <0x…> (a hudson.plugins.sshslaves.SSHLauncher)
	at hudson.plugins.sshslaves.SSHLauncher.afterDisconnect(SSHLauncher.java:925)
	at hudson.slaves.SlaveComputer$3.run(SlaveComputer.java:754)
	at …
"Computer.threadPoolForRemoting [#…]" … in Object.wait() …
   java.lang.Thread.State: WAITING (on object monitor)
	at java.lang.Object.wait(Native Method)
	at java.lang.Object.wait(Object.java:502)
	at com.trilead.ssh2.channel.ChannelManager.waitUntilChannelOpen(ChannelManager.java:110)
	- locked <0x…> (a com.trilead.ssh2.channel.Channel)
	at com.trilead.ssh2.channel.ChannelManager.openSessionChannel(ChannelManager.java:574)
	at com.trilead.ssh2.Session.<init>(Session.java:42)
	at com.trilead.ssh2.Connection.openSession(Connection.java:1059)
	- locked <0x…> (a com.trilead.ssh2.Connection)
	at com.trilead.ssh2.Connection.exec(Connection.java:1450)
	at hudson.plugins.sshslaves.SSHLauncher$3.run(SSHLauncher.java:969)
	at …
```

PR #180, opened 2020-04-08
Improved buildConfiguration

jglick:
I noticed that #178 was not publishing binaries due to https://github.com/jenkins-infra/pipeline-library/blob/c7acbe5bf68da0f1ad00d63d40a38e9dc59d6e6d/vars/buildPlugin.groovy#L58. While I am here, cleaning up a bit and aligning with https://github.com/jenkins-infra/pipeline-library/pull/120.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,13 +1,9 @@
 //def buildConfiguration = buildPlugin.recommendedConfigurations()
 
 def buildConfiguration = [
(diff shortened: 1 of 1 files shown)


jglick:
There we go: https://ci.jenkins.io/job/Plugins/job/ssh-slaves-plugin/job/PR-180/1/artifact/

PR #183, opened 2020-08-27
Fix issue with Java installed into a path with spaces

slide:
This can happen on Windows most often with the existing installers (AdoptOpenJDK, Oracle JDK, etc).

I don't think this needs tests, but would be happy to look into it if maintainers think otherwise.

### Submitter checklist

- [X] Appropriate autotests or explanation to why this change has no tests

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/JavaVersionChecker.java
+++ b/src/main/java/hudson/plugins/sshslaves/JavaVersionChecker.java
@@ -88,11 +88,7 @@ public class JavaVersionChecker {
         StringWriter output = new StringWriter();   // record output from Java
 
         ByteArrayOutputStream out = new ByteArrayOutputStream();
-        // if the command has a space, we need to quote it
-        if(javaCommand.contains(" ")) {
-            javaCommand = "\"" + javaCommand + "\"";
-        }
-        connection.exec(javaCommand + " "+ jvmOptions + " -version",out);
+        connection.exec("\"" + javaCommand + "\" "+ jvmOptions + " -version",out);
         //TODO: Seems we need to retrieve the encoding from the connection destination
         BufferedReader r = new BufferedReader(new InputStreamReader(
                 new ByteArrayInputStream(out.toByteArray()), Charset.defaultCharset()));
--- a/src/main/java/hudson/plugins/sshslaves/JavaVersionChecker.java
+++ b/src/main/java/hudson/plugins/sshslaves/JavaVersionChecker.java
@@ -88,6 +88,10 @@ public class JavaVersionChecker {
         StringWriter output = new StringWriter();   // record output from Java
 
         ByteArrayOutputStream out = new ByteArrayOutputStream();
+        // if the command has a space, we need to quote it
+        if(javaCommand.contains(" ")) {
+            javaCommand = "\"" + javaCommand + "\"";
+        }
         connection.exec(javaCommand + " "+ jvmOptions + " -version",out);
         //TODO: Seems we need to retrieve the encoding from the connection destination
         BufferedReader r = new BufferedReader(new InputStreamReader(


oleg-nenashev (approved):

slide:
> It'd be more appropriate to use proper process APIs to avoid the need to manually quote things here, but this should fix the issue at least.

What API are you referring to (I am not a Java developer in my day job). The `connection` object is an SSH connection, this is executing the process remotely via SSH, so I am not sure how process API's would help in this case, unless there is a class that handles building a full command that can then be sent to the `connection.exec` method.

jvz:
In this case, the "proper API" would have to be some Jenkins (or more likely Apache Commons or something) util class that performs the quoting for you. I wrote something like that for encoding Windows commands a while ago so that we could "properly" test file junctions (since the only way to create those is through a cmd.exe builtin, so yay for double-escaping), and I'd hope something like that also exists for sh-compatible escaping, too.

slide:
Three is https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/util/ArgumentListBuilder.java, but it only takes into account spaces in paths, not other special characters. The other end of the command may be either Windows or UNIX, so an sh-compatible escape won't work in this case.

src/main/java/hudson/plugins/sshslaves/JavaVersionChecker.java
```diff
@@ -88,6 +88,10 @@ private String checkJavaVersion(TaskListener listener, String javaCommand) throw
         StringWriter output = new StringWriter();   // record output from Java
 
         ByteArrayOutputStream out = new ByteArrayOutputStream();
+        // if the command has a space, we need to quote it
+        if(javaCommand.contains(" ")) {
+            javaCommand = "\"" + javaCommand + "\"";
+        }
         connection.exec(javaCommand + " "+ jvmOptions + " -version",out);
```
kuisathaverat, line 95: Using Path we can avoid other kinds of issues in the path
you have to add these two imports 

```
java.nio.file.Paths
java.nio.file.Path
```

and apply this patch
```suggestion
        Path javaPath = Paths.get(javaCommand);
        connection.exec(javaPath.toString() + " "+ jvmOptions + " -version",out);
```

slide: That does not add quotes around the path if it contains spaces.

kuisathaverat: shit, it is true

kuisathaverat, line 95: another option is to put the Java path between quotes it should work on windows and Unix

```suggestion
        connection.exec("\"" + javaCommand + "\" "+ jvmOptions + " -version",out);
```

kuisathaverat:
I found it (and we can not use it), Apache commons has a library to compose and execute commands with cross-OS support https://commons.apache.org/proper/commons-exec/commandline.html but it forces you to quote arguments it then contains whitespaces so it does not fix the issue, also the command generation will be run on the Jenkins instance so the OS detection will use the Jenkins instance OS, and because of that the separators and other stuff will not valid if the Agent has different OS.

slide:
Always quoting may be the way to go instead of checking for specific characters. Thoughts?

jvz:
Quoting it will work for most situations. If people put their Java installation in a directory like `/usr/$java/bin` or `/usr/"java"/bin` or similar special characters that would otherwise also need to be escaped.

kuisathaverat:
I think is fine the Java check is only used if you do not set the Java Path https://github.com/jenkinsci/ssh-slaves-plugin/blob/64116586a7cea94d520d29906b64160d9008ce88/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java#L446-L453

Set the Java Path with the sull path between quotes is a configuration workaround also.

kuisathaverat (approved):

kuisathaverat:
I will merge in a couple of days if there is not any other concern.

slide:
> I will merge in a couple of days if there is not any other concern.

Do you want the always quoted version?

kuisathaverat:
I preferred always quoted the logic is simpler, but you decided is your PR

merlindam:
Hi, 
got issue with this fix in our configuration ( Windows node using cygwin ssh server, and environment JAVA_HOME set to : C:\Program Files\Java\jre8 but our need is not to use this to connect to node)
Before your change the code was unable to get the version of the provided java : 
```
Couldn't figure out the Java version of c:\Program Files\Java\jre8/bin/java
bash: c:Program: command not found
```
So was not used when starting agent.
But now with the new code the version is retrieve but the start of the agent failed 
```
12/09/20 11:54:29] [SSH] Starting agent process: cd "e:\jenkins-p" && C:/Program Files/Java/jre8/bin/java  -jar remoting.jar -workDir e:\jenkins-p -jar-cache e:\jenkins-p/remoting/jarCache
bash: C:/Program: No such file or directory
```
If I'm correct the issue is just with the last change you did in this PR when you decided to always quoted, you should have kept the change to the java _commandJava_ variable else the correct command return is not the one needed.

merlindam (commented):
I disagree with this change.
If need to always quote, you should update javaCommand in order that the correct command to use be return. Check more details in my PR comment

PR #184, opened 2020-10-08
Enable dependabot to automatically get dependency updates

MRamonLeon:
It would be great to have **Dependabot** configured to get dependencies updates filed automatically. This will help avoiding bugs or security vulnerabilities from old dependencies.

Some helpful documentation about Dependabot to fine adjust it if needed:

* Keeping your dependencies updated automatically: https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/keeping-your-dependencies-updated-automatically
* Avoiding security vulnerabilities: https://docs.github.com/en/free-pro-team@latest/github/managing-security-vulnerabilities/configuring-github-dependabot-security-updates
* Fine adjusting the configuration: https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "maven"
+    directory: "/"
+    schedule:
+      interval: "weekly"


MRamonLeon:
cc: @oleg-nenashev @jglick

oleg-nenashev (dismissed):
This pattern would lead dependabot to suggest Jenkins core dependency upgrades. Please add it to the ignore list

jglick:
> This pattern would lead dependabot to suggest Jenkins core dependency upgrades.

No, it does not.

jglick (approved):

kuisathaverat (approved):

kuisathaverat:
as soon as @oleg-nenashev approve it I will merge it

oleg-nenashev (approved):
> No, it does not.

Hmm, really? Not clear why, but I do not want to block it. can be fixed later

jglick:
> Not clear why

I think because while the _property_ is defined in the visible POM, the _dependency_ is defined in the parent, and Dependabot does not try to resolve the effective POM—it does not do the precise analysis that

```bash
mvn versions:display-dependency-updates
```

would.

jglick:
Thus our stock recommendation: https://github.com/jenkinsci/archetypes/blob/61e835797f4c8318336b37aaa107eedac62dd6ce/common-files/.github/dependabot.yml

oleg-nenashev:
> I think because while the *property* is defined in the visible POM, the
*dependency* is defined in the parent, and Dependabot does not try to
resolve the effective POM—it does not do the precise analysis that

It was definitely doing it before for Jenkins core. Maybe there was a
regression in Dependabot at some point, I have noticed some degradation in
"GitHub-native dependabot" for other projects. Some features are still not
available in the new edition.

BR, Oleg

jglick:
BTW I strongly recommend switching to https://github.com/jenkinsci/bom#usage

kuisathaverat:
@jglick when I bump the core version I will do it, as I did for [trilead-api](https://github.com/jenkinsci/trilead-api-plugin/commit/f4a23faaa144d25a09c3d3d9e13224fd3005a532)

PR #199, opened 2020-12-09
[JENKINS-64404] Additional fix for Java installed into a path with spaces

merlindam:
Those changes complete :
commit 9646cef228485f3bdf2cbec34b0dc7d0f3fa3d0e
Fix issue with Java installed into a path with spaces (#183)

See [JENKINS-64404](https://issues.jenkins-ci.org/browse/JENKINS-64404).

### Submitter checklist

- [ x ] JIRA issue is well described
- [ x ] Appropriate autotests or explanation to why this change has no tests ( PR build validate on our preprod server)

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/JavaVersionChecker.java
+++ b/src/main/java/hudson/plugins/sshslaves/JavaVersionChecker.java
@@ -88,7 +88,9 @@ public class JavaVersionChecker {
         StringWriter output = new StringWriter();   // record output from Java
 
         ByteArrayOutputStream out = new ByteArrayOutputStream();
-        connection.exec("\"" + javaCommand + "\" "+ jvmOptions + " -version",out);
+        //Always quoted java command in order to handle spaces in java path
+        javaCommand = "\"" + javaCommand + "\"";
+        connection.exec(javaCommand + " "+ jvmOptions + " -version",out);
         //TODO: Seems we need to retrieve the encoding from the connection destination
         BufferedReader r = new BufferedReader(new InputStreamReader(
                 new ByteArrayInputStream(out.toByteArray()), Charset.defaultCharset()));


kuisathaverat:
```
connection.exec("\"" + javaCommand + "\" "+ jvmOptions + " -version",out);
```

and 

```
        javaCommand = "\"" + javaCommand + "\"";
        connection.exec(javaCommand + " "+ jvmOptions + " -version",out);
```

have exactly the same output

merlindam:
> 
> 
> ```
> connection.exec("\"" + javaCommand + "\" "+ jvmOptions + " -version",out);
> ```
> 
> and
> 
> ```
>         javaCommand = "\"" + javaCommand + "\"";
>         connection.exec(javaCommand + " "+ jvmOptions + " -version",out);
> ```
> 
> have exactly the same output


Sure completely.
But the change here is to be sure to modify `javaCommand` as it is used afterwards in the function see below why.

In the following of the function the output of the -version command AND the `javaCommand` is passed to a function that checked the output 
https://github.com/jenkinsci/ssh-slaves-plugin/blob/c0ae5272c7763d5747c20da26a5f014d6bb39bbc/src/main/java/hudson/plugins/sshslaves/JavaVersionChecker.java#L95
and if correct the contents of `javaCommand` is return as a result.
https://github.com/jenkinsci/ssh-slaves-plugin/blob/c0ae5272c7763d5747c20da26a5f014d6bb39bbc/src/main/java/hudson/plugins/sshslaves/JavaVersionChecker.java#L143

Then `resolveJava` function return this path 
https://github.com/jenkinsci/ssh-slaves-plugin/blob/c0ae5272c7763d5747c20da26a5f014d6bb39bbc/src/main/java/hudson/plugins/sshslaves/JavaVersionChecker.java#L75

So then in `launch` function : 
https://github.com/jenkinsci/ssh-slaves-plugin/blob/c0ae5272c7763d5747c20da26a5f014d6bb39bbc/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java#L408
the initial `javaCommand` ends up in `java` variable : 
https://github.com/jenkinsci/ssh-slaves-plugin/blob/c0ae5272c7763d5747c20da26a5f014d6bb39bbc/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java#L452
that is used in the function that start the agent:
https://github.com/jenkinsci/ssh-slaves-plugin/blob/c0ae5272c7763d5747c20da26a5f014d6bb39bbc/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java#L457

So with the current state as I reported in [JENKINS-64404](https://issues.jenkins.io/browse/JENKINS-64404) the issue is for our case that the launch agent failed because of missing quotes: 
```
[12/09/20 11:54:29] [SSH] Starting agent process: cd "e:\jenkins-p" && C:/Program Files/Java/jre8/bin/java -jar remoting.jar -workDir e:\jenkins-p -jar-cache e:\jenkins-p/remoting/jarCache
bash: C:/Program: No such file or directory
```

merlindam:
Change successfully validated on our prepod jenkins server, for the failing configuration
```
[12/10/20 16:28:38] [SSH] Checking java version of e:\jenkins-p/jdk/bin/java
Couldn't figure out the Java version of "e:\jenkins-p/jdk/bin/java"
bash: e:\jenkins-p/jdk/bin/java: No such file or directory

[12/10/20 16:28:38] [SSH] Checking java version of C:/Program Files/Java/jre8/bin/java
[12/10/20 16:28:38] [SSH] "C:/Program Files/Java/jre8/bin/java" -version returned 1.8.0_101.
[12/10/20 16:28:38] [SSH] Starting sftp client.
[12/10/20 16:28:38] [SSH] Copying latest remoting.jar...
Source agent hash is D866F0B482DB94F38E49B26B465D5DB5. Installed agent hash is D866F0B482DB94F38E49B26B465D5DB5
Verified agent jar. No update is necessary.
Expanded the channel window size to 4MB
[12/10/20 16:29:02] [SSH] Starting agent process: cd "e:\jenkins-p" && "C:/Program Files/Java/jre8/bin/java"  -jar remoting.jar -workDir e:\jenkins-p -jar-cache e:\jenkins-p/remoting/jarCache
Dec 10, 2020 4:29:02 PM org.jenkinsci.remoting.engine.WorkDirManager initializeWorkDir
INFO: Using e:jenkins-p\remoting as a remoting work directory
Dec 10, 2020 4:29:02 PM org.jenkinsci.remoting.engine.WorkDirManager setupLogging
INFO: Both error and output logs will be printed to e:jenkins-p\remoting
<===[JENKINS REMOTING CAPACITY]===>channel started
Remoting version: 4.5
This is a Windows agent
[StartupTrigger] - Scanning jobs for node test-w10-64-test-template-sdkt037
Agent successfully connected and online
```
And this is also still working on our others nodes.
Note that we have a good variety of configuration ( mac, linux with various distribution and version, windows,...)

kuisathaverat:
well, change a string that is passed as a parameter, and is loaded from another place, sounds like a bad idea, it hides the change on the string in a function is not related at all to change this value. I will try to fin another way. 
The recommended way to take the java command for the remoting process (remoting.jar) is from the PATH, if you need a different Java version you should use the javaPath advanced setting.
I was thinking to deprecate the JavaVersionChecker class a long time ago, Java is a requirement to run the remoting process so it should be configured on the agent environment or in the agent configuration, anything else is trying to resolve a configuration problem, it is odd, error prone, and never will cover all the alternatives.

kuisathaverat:
superseded by https://github.com/jenkinsci/ssh-slaves-plugin/pull/202

merlindam:
I fully agree that change in PR #202 are much better than the one's I tired.
The new PR successfully replace this one, I will close this PR.

PR #209, opened 2020-12-27
Add /opt/java/openjdk as a potential Java source for jenkins/ssh-agent AdoptOpenJDK Alpine images

oleg-nenashev:
Jus a minor tweak which helps the instance to discover Java when using the new AdoptOpenJDK packages for alpine. https://github.com/jenkinsci/docker-ssh-agent/blob/master/8/alpine/Dockerfile 

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/DefaultJavaProvider.java
+++ b/src/main/java/hudson/plugins/sshslaves/DefaultJavaProvider.java
@@ -82,7 +82,7 @@ public class DefaultJavaProvider extends JavaProvider {
                                    "/usr/java/latest/bin/java",
                                    "/usr/local/bin/java",
                                    "/usr/local/java/bin/java",
-                                   "/opt/java/openjdk"));
+                                   "/opt/java/openjdk/bin/java"));
         return javas;
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/sshslaves/DefaultJavaProvider.java
```diff
@@ -81,7 +81,8 @@
                                    "/usr/java/default/bin/java",
                                    "/usr/java/latest/bin/java",
                                    "/usr/local/bin/java",
-                                   "/usr/local/java/bin/java"));
+                                   "/usr/local/java/bin/java",
+                                   "/opt/java/openjdk"));
```
oleg-nenashev, line 85:
```suggestion
                                   "/opt/java/openjdk/bin/java"));
```

MarkEWaite (approved):
Confirmed that `/opt/java/openjdk/bin/java` is installed in `jenkins/ssh-agent:3.0.0-alpine` and reports Java version 1.8.0_262.

kuisathaverat:
we cannot add all the path that exists n the world, the JVM should be on the PATH because is a requirement, if not you should pass the java path setting. Here we are trying to resolve an issue with paths in the Docker image with a sticker in another place. I've deprecated the java check class a few days a do and I will remove it in a year or so.
The proper solution for the Docker images is to set the environment correctly,  I have done it for months in my test environments https://github.com/kuisathaverat/jenkins-issues/blob/master/kex-algorithms/ssh-agent-dsa/Dockerfile#L18

kuisathaverat:
The issue is in the Docker image not in the plugin and the solution is to fix the Docker image by fixing https://github.com/jenkinsci/docker-ssh-agent/issues/33, you can do it by adding Java to the Path or by creating a symbolic link on /usr/bin/java. The proper fix could be https://github.com/jenkinsci/docker-ssh-agent/pull/74

kuisathaverat:
resolved by https://github.com/jenkinsci/docker-ssh-agent/pull/78

PR #215, opened 2021-03-12
Update baseline to 2.282, fixes PCT (detached ssh plugin)

timja:
See https://github.com/jenkinsci/bom/pull/449 and https://github.com/jenkinsci/jenkins/pull/5049

Tests fail on 2.282 without this change

We will need a release for bom please

cc @kuisathaverat

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -27,9 +27,9 @@
   <properties>
     <revision>1.31.6</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.222.4</jenkins.version>
+    <jenkins.version>2.282</jenkins.version>
     <java.level>8</java.level>
-    <configuration-as-code.version>1.46</configuration-as-code.version>
+    <hpi.compatibleSinceVersion>1.31.0</hpi.compatibleSinceVersion>
   </properties>
 
   <developers>
@@ -102,8 +102,8 @@
     <dependencies>
       <dependency>
         <groupId>io.jenkins.tools.bom</groupId>
-        <artifactId>bom-2.204.x</artifactId>
-        <version>18</version>
+        <artifactId>bom-2.277.x</artifactId>
+        <version>26</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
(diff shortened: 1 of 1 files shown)


kuisathaverat (approved):

timja:
could you release please @kuisathaverat ?

kuisathaverat:
Sure, tomorrow morning I will find some time

mfriedenhagen:
While I am always a friend of timely updates, the bump of the required core version of Jenkins (and especially to a non LTS one) was unexpected when only the patch/micro version of the plugin changed.

kuisathaverat:
Jenkins core version bump is a minor change I made every six months, I only support 2 of latest LTS. It is true that this is a little weird but it is because of the nature of the change, we have removed the sshd-module from the Jenkins core and this change is only on the weekly for the moment. But yes maybe I could make a new minor (1.32) and keep the branch of the 1.31 for urgent patches. In any case, probably in the next 1-2 month, I will release a major release with new features and I do not expect to release patches.

andham:
The requirement of Jenkins v2.282 will prevent all of us using Jenkins LTS to upgrade for at least 4 months, which is very bad. This change should not have been in a bugfix IMHO but mandates a minor bump.

timja:
If there's something critical that is released i would expect it to be backported to a version that supports the LTS.

If it's not critical and it's something you want then you can:
1. ask for a backport
2. make a custom build.

Not sure why this is 'very bad'

You are currently on the latest supported version of your release line

Due to changes in Jenkins core this plugin needed to move to a newer minimum version.

andham:
@timja Fair point. Appologize for the strong wording. I appreciate the required code change for users on weekly, but there is a conflict here with LTS users. This requirement bump in a bugfix release was suprising for me as a LTS user.

mfriedenhagen:
Slightly off-topic: maybe plugins should even bump their major version when they use a new core? Not sure whether there is a general recommendation for Jenkins plugins. E.g. for Maven plugins I would expect a new major version when a new minor of Maven would be required.

Anyways, thanks @timja and @kuisathaverat for your work.

PR #253, opened 2021-09-06
update System.out.println to listener.getLogger().println

mamh2021:
update System.out.println to listener.getLogger().println in src/main/java/hudson/plugins/sshslaves/SSHLauncher.java 

### Submitter checklist


kuisathaverat:
The `System.out` output messages are sent to the Jenkins logs and the `logger` messages are sent to the Agent logs, so the behavior is correct, this PR is not need.

mamh2021:
Why are these three lines use System.out？
print messages to jenkins logs why not 
use `private static final Logger LOGGER = Logger.getLogger(SSHLauncher.class.getName()); ` ???

PR #269, opened 2022-01-21
feat: enable continuous delivery

kuisathaverat:
This comes with a new version number format, see https://www.jenkins.io/jep/229

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
   </parent>
 
   <artifactId>ssh-slaves</artifactId>
-  <version>${revision}${changelist}</version>
+  <version>${revision}-${changelist}</version>
   <packaging>hpi</packaging>
 
   <name>SSH Build Agents plugin</name>
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,4 +3,8 @@ updates:
   - package-ecosystem: "maven"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "daily"
+  - package-ecosystem: "github-actions"
+    directory: /
+    schedule:
+      interval: "daily"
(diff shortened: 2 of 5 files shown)


pom.xml
```diff
@@ -9,7 +9,7 @@
   </parent>
 
   <artifactId>ssh-slaves</artifactId>
-  <version>${revision}${changelist}</version>
+  <version>${revision}-${changelist}</version>
```
jtnord, line 12: @kuisathaverat this will produce very strange version numbers (will all be 1.33.1-DDD-hash`

the `DDD-hash` is all going to be sorted lexically and everything will always be some version of `1.33.1`  (which is normal only for an API plugin delivering a library at version `1.33.1`  (e.g. jackson2-api or caffeine-api)

In other places where we have wanted to keep a major/minor component before the changelist we have just hard coded it and changed the changelist.format as `%d-v%s`   and set the version to be `<version>1.${changelist}</version>`

kuisathaverat: yeah, I saw the version generated `1.33.1-805.vb_a_0d6a_21f322)` not nice, if we use `<version>1.${changelist}</version>` it will be `1.805.vb_a_0d6a_21f322` that it is odd also, I was expecting something like `1.33.850.a2343a_43a24` that is the incrementals format.


The change list format is the one suggested by [Setting up automated plugin release](https://www.jenkins.io/doc/developer/publishing/releasing-cd/) `-Dchangelist.format=%d.v%s`

kuisathaverat: that PR have exactly the `<version>` (change the dot with a dash) I have chosen https://github.com/jenkins-infra/jenkins.io/pull/4933/files#diff-be0d6b1d91f3683c5535db323262cc5a2dfb587e23baac80b55f1260259de8cbR151-R170

kuisathaverat: I see it changes also the revision so it is not the same

kuisathaverat: I have opened https://github.com/jenkinsci/ssh-slaves-plugin/pull/278

PR #273, opened 2022-02-26
fixup System.out.println to listener.getLogger().println

mamh2021:
fixup System.out.println to listener.getLogger().println

same as in file src/main/java/hudson/plugins/sshslaves/SSHLauncher.java line 493, print the messages to listener.getLogger()
https://github.com/jenkinsci/ssh-slaves-plugin/blob/main/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java#L493

I think replace System.out.println to listener.getLogger()  is better.

or maybe  replace System.out.println to  Logger LOGGER = Logger.getLogger(SSHLauncher.class.getName()); 
The System.out output messages are sent to the Jenkins logs maybe use LOGGER is best way.

Signed-off-by: bright.ma rmsh06 <bright.ma@blackshark.com>

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -493,16 +493,16 @@ public class SSHLauncher extends ComputerLauncher {
                     listener.getLogger().println(Messages.SSHLauncher_launchCanceled());
                 }
                 if (!res) {
-                    System.out.println(Messages.SSHLauncher_LaunchFailedDuration(getTimestamp(),
+                    listener.getLogger().println(Messages.SSHLauncher_LaunchFailedDuration(getTimestamp(),
                             nodeName, host, duration));
                     listener.getLogger().println(getTimestamp() + " Launch failed - cleaning up connection");
                     cleanupConnection(listener);
                 } else {
-                    System.out.println(Messages.SSHLauncher_LaunchCompletedDuration(getTimestamp(),
+                    listener.getLogger().println(Messages.SSHLauncher_LaunchCompletedDuration(getTimestamp(),
                             nodeName, host, duration));
                 }
             } catch (InterruptedException e) {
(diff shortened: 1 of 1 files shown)


kuisathaverat:
again https://github.com/jenkinsci/ssh-slaves-plugin/pull/253

mamh2021:
@kuisathaverat 
Can you explain why these lines are used "System.out.println"
why not same as SSHLauncher.java line 493 ?

kuisathaverat:
as I said in the previous PR, the `stdout` goes to the Agent longs, and the `logger` goes to the Jenkins logs, are two completely different destinations.

mamh2021:
yes, they are different destinations，whichever works fine ,

but to Agent logs why not ` LOGGER = Logger.getLogger(SSHLauncher.class.getName());` ???

kuisathaverat:
Did you even test your code? Did you even notice that you are removing those messages from the Agent log console and moving them to the Jenkins controller logs? this case is not the case described in the StackOverflow link you pass, in this case, is not a bad practice, the stdin of this process is redirected as Agent log output, it is how it is designed and it is how it works.

mamh2021:
Of course I've tested my code, They work very well  on my product jenkins. these logs all print to "computer/my slave name/log" on the webui,  This is the effect I want . I didn't want to see these logs print in "/var/log/jenkins/jenkins.log" on my jenkins master server.

Your statement is not well explained

kuisathaverat:
as I said `stderr` and `LOGGER` are different logs for different purposes.

for more details :
* [Remoting Architecture](https://github.com/hudson/www/blob/master/docs/HudsonArch-Remoting.pdf)
* [Remoting project](https://www.jenkins.io/projects/remoting/)
* [SlaveComputer.setChannel](https://javadoc.jenkins.io/hudson/slaves/SlaveComputer.html#setChannel(java.io.InputStream,java.io.OutputStream,hudson.model.TaskListener,hudson.remoting.Channel.Listener))

PR #274, opened 2022-02-28
Changes monitor category

alecharp:
Since Jenkins 2.267, there are 2 different category of monitors: security and non security ones.
In this plugin, the administrative monitor is enabled when an agent is connected using SSH keys but no key validation strategy is enabled is the connection configuration.
As this could lead to issues with those agent, it would be preferable that this monitor to be considered as a security one.

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/verifiers/MissingVerificationStrategyAdministrativeMonitor.java
+++ b/src/main/java/hudson/plugins/sshslaves/verifiers/MissingVerificationStrategyAdministrativeMonitor.java
@@ -65,6 +65,11 @@ public class MissingVerificationStrategyAdministrativeMonitor extends Administra
     }
 
     public String getAgentNames() {
-      return agentNames != null ? agentNames.toString() : "";
+        return agentNames != null ? agentNames.toString() : "";
+    }
+
+    @Override
+    public boolean isSecurity() {
+        return true;
     }
 }


alecharp:
@kuisathaverat do you think you could review this? Thanks

jtnord (approved):

raul-arabaolaza (approved):

kuisathaverat (approved):

alecharp:
@kuisathaverat do you think you could trigger a release with those changes? Thanks.

kuisathaverat:
🤔 it should be triggered, I just configure the plugin to automatically release user affected changes https://www.jenkins.io/doc/developer/publishing/releasing-cd/ 
Checking the commits the CI did not pass in the latest commits, I will wait for the latest commit to see if it pass, if not, and depending on the issue, I would release.

alecharp:
I think for it to be automatic, any PR merged needs to be labeled with an "interesting" label. `dependency` is not one of them. But normally, in the actions, you can trigger the release manually.

kuisathaverat:
This PR has the label `enhancement`, and is in the changes since the latest release it should trigger the release

kuisathaverat:
it is not, I guess it only checks the current commit, I thought it checked the diffs since the latest release
https://github.com/jenkinsci/ssh-slaves-plugin/actions/runs/1927313100

alecharp:
the check ci status fails because there are tests issues on the primary branch.

kuisathaverat:
>the check ci status fails because there are tests issues on the primary branch.
See https://github.com/jenkinsci/ssh-slaves-plugin/runs/5377315062

those are random timeouts on the Jenkins CI it takes more than 180 seconds to provision a Docker container sometimes. I have tricked the CD workflow and there is a new release

PR #283, opened 2022-04-08
[JENKINS-68195] Duplicated help icon for Host Key verification

kuisathaverat:
See [JENKINS-68195](https://issues.jenkins-ci.org/browse/JENKINS-68195).

### Submitter checklist

- [X] JIRA issue is well described
- [X] Appropriate autotests or explanation to why this change has no tests

After some improvements in the `f:dropdownDescriptorSelector` and the UI present in Jenkins Core 2.242 I need to change a couple of things in some jelly files. I have removed a workaround to show the help link on `f:dropdownDescriptorSelector` and I have to put the `f:dropdownDescriptorSelector` inside an `<f:entry>` to avoid showing the help button in the end of the right side of the browser. This is the visual result of the changes:
Before:
Screenshot 2022-04-08 at 11 41 18
After:
Screenshot 2022-04-08 at 11 17 03

Related to https://github.com/jenkinsci/jenkins/pull/5417

## Note
This PR cannot be merged before Jenkins Core 2.42 (or upper version) is released as LTS because it breaks the help links in previous versions.

PR #304, opened 2022-06-09
Fix deadlock when afterDisconnect is called during launch()

tomekjarosik:
Deadlock happens between:
(a) [synchronized void tearDownConnection](https://github.com/jenkinsci/ssh-slaves-plugin/blob/main/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java#L953)
(b) [synchronized (this) {](https://github.com/jenkinsci/ssh-slaves-plugin/blob/main/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java#L413)

Launch() is stuck here: [results = srv.invokeAll(callables);](https://github.com/jenkinsci/ssh-slaves-plugin/blob/main/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java#L486) ,
and `callable` is stuck waiting here: [lock.wait(); // wait until the writer gives us something](https://github.com/jenkinsci/trilead-ssh2/blob/8e7118e74deacd7ef044df9076fcb9b95280c163/src/com/trilead/ssh2/channel/FifoBuffer.java#L212)
After some time `afterDisconnect` is called which initiates tear down process. Teardown process calls [launcherExecutorService.shutdown()](https://github.com/jenkinsci/ssh-slaves-plugin/blob/main/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java#L941) and continues to (a).
But because there is no writer writing to FifoBuffer, launch() never finishes and is stuck in (b), and because `shutdown()` only prevents new tasks being submitted (see [docs](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ExecutorService.html)) there is no one to interrupt lock.wait() in FifoBuffer, thus deadlock. The solutions is to use recommended way of shutting down ExecutorService from docs.

See [JENKINS-68656](https://issues.jenkins.io/browse/JENKINS-68656).

### Submitter checklist

- [x] JIRA issue is well described
- [x] No new tests, because there are existing tests that sometimes hang (see one example mentioned in the comments section)
- [x] overnight test launching ephemeral agent around ~800 times

### DIFF
--- a/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
+++ b/src/main/java/hudson/plugins/sshslaves/SSHLauncher.java
@@ -505,11 +505,7 @@ public class SSHLauncher extends ComputerLauncher {
                 System.out.println(Messages.SSHLauncher_LaunchFailed(getTimestamp(),
                         nodeName, host));
             } finally {
-                ExecutorService srv = launcherExecutorService;
-                if (srv != null) {
-                    srv.shutdownNow();
-                    launcherExecutorService = null;
-                }
+              shutdownAndAwaitTerminationOfLauncher();
             }
         }
         if (node != null && getTrackCredentials()) {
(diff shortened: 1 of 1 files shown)


kuisathaverat:
>Also, the mentioned deadlock happened here too: https://github.com/jenkinsci/ssh-slaves-plugin/pull/283/checks?check_run_id=5979901734 (see "Standard error" logs there)

In this PR the Docker agent is not provisioned in time, the issue is in the Docker agent not in the plugin, and the test failed by timeout.

kuisathaverat:
Did you test this PR and it resolves the issue?

You have an incremental plugin  to tests it https://github.com/jenkinsci/ssh-slaves-plugin/pull/304/checks?check_run_id=6811122752

PR #322, opened 2022-07-18
increase jenkinsrule timeout for flaky tests

Pldi23:
SshLauncherTest#trackCredentialsWithUsernameAndPrivateKey and SshLauncherTest#trackCredentialsWithUsernameAndPassword are flaky.

Tests itself are good, but from time to time they hit JenkinsRule's timeout (180 sec), so I suggest to increase timeout to 600 sec as some of tests in other plugin do [(search)](https://github.com/search?q=org%3Ajenkinsci+WithTimeout+600&type=code)

### Submitter checklist

### DIFF
--- a/src/test/java/hudson/plugins/sshslaves/SSHLauncherTest.java
+++ b/src/test/java/hudson/plugins/sshslaves/SSHLauncherTest.java
@@ -66,6 +66,8 @@ import com.cloudbees.plugins.credentials.SystemCredentialsProvider;
 import com.cloudbees.plugins.credentials.domains.Domain;
 import com.cloudbees.plugins.credentials.domains.HostnamePortSpecification;
 import com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl;
+import org.jvnet.hudson.test.recipes.WithTimeout;
+
 import static hudson.plugins.sshslaves.SSHLauncher.JAR_CACHE_DIR;
 import static hudson.plugins.sshslaves.SSHLauncher.JAR_CACHE_PARAM;
 import static hudson.plugins.sshslaves.SSHLauncher.WORK_DIR_PARAM;
@@ -239,6 +241,7 @@ public class SSHLauncherTest {
   }
 
   @Issue("JENKINS-38832")
+  @WithTimeout(600)
   @Test
   public void trackCredentialsWithUsernameAndPassword() throws Exception {
     UsernamePasswordCredentialsImpl credentials = new UsernamePasswordCredentialsImpl(CredentialsScope.SYSTEM, "dummyCredentialId", null, "user", "pass");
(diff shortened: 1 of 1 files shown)


kuisathaverat:
I did not see those tests failed in [ci.jenkins.io](https://ci.jenkins.io), Could you give more details about where you see those failures?

Pldi23:
> I did not see those tests failed in [ci.jenkins.io](https://ci.jenkins.io), Could you give more details about where you see those failures?

Cloudbees CI fails intermittently when running this test, I'will attach the log soon

kuisathaverat:
>Cloudbees CI fails intermittently when running this test, I'will attach the log soon

It fails on Cloudbees CI  and not in [ci.jenkins.io](https://ci.jenkins.io/) looks like an issue on Cloudbees CI not in the tests. when a test is flaky, it is flaky everywhere not only in one CI/environment.

kuisathaverat:
Checking the logs, the agent is created and the SSH agent is trying to reconnect two tomes so, for some reason the Jenkins is not responsible enough in your CI, it looks like resources issues on the Agent used to run the test.

Pldi23:
> > Cloudbees CI fails intermittently when running this test, I'will attach the log soon
> 
> It fails on Cloudbees CI and not in [ci.jenkins.io](https://ci.jenkins.io/) looks like an issue on Cloudbees CI not in the tests. when a test is flaky, it is flaky everywhere not only in one CI/environment.

Hm, I tried to run this test in the IntelliJ and it failed on first attempt with the same log (next 9 attempts was fine), it is not looks like Cloudbees CI issue.

kuisathaverat:
Which commit are you using? Which JDK are you using? In the last three days ci.jenkins.io has running the test 3 times, in the last 11 days there is no issues related to those tests

Pldi23:
I'm using latest main branch, openjdk version "1.8.0_292". IMO 11 runs is not representative, in our CI it fails once in hundreds of runs (other note: when I running locally on system with low resources it fails each time)

kuisathaverat:
I have ran it on my local three times without issues, I am using JDK 11, JDK 8 is not recommended in the Jenkins core 2.346.2 version (since https://www.jenkins.io/changelog-stable/#v2.332.2) I am no longer compile on JDK 8 since this plugin is updated to the latest BOM and does not make sense compile on JDK 8 anymore.
However, checking the execution of the test should take about 150-160 seconds (you can check the execution in the Agent logs in the Jenkins instantiated by the test), so it is possible to make the test faster by configuring less wait between reconnections and less timeout of connection. This change will resolve your issue and improve the test execution instead of making it wait longer.

kuisathaverat:
I am making a PR with the changes

Pldi23:
> it is possible to make the test faster by configuring less wait between reconnections and less timeout of connection. This change will resolve your issue and improve the test execution instead of making it wait longer.

I agree, sounds much better

kuisathaverat:
superseded by https://github.com/jenkinsci/ssh-slaves-plugin/pull/323

PR #323, opened 2022-07-18
fix: change the test connection parameters to fail faster

kuisathaverat:
supersede https://github.com/jenkinsci/ssh-slaves-plugin/pull/322


Pldi23:
Had a 200 rans on a low resources VM, fix works perfectly.

PR #368, opened 2022-12-23
Fix start from workspace on drive D:

kgermanov:
When setup workspace for ex D:\workspace it failed, because on windows the command
`cd D:\workspace` 
does not work when current_dir on drive C:
It need to add option /D :
`cd /D D:\workspace` 
But more portable dont change working directory - intead of this simple start "remoting.jar" by absolute path.


kuisathaverat:
This will create all the remoting folders in the execution path not in the workdir so it is not a valid solution, check the troubleshooting guide for a configuration solution https://github.com/jenkinsci/ssh-slaves-plugin/blob/main/doc/TROUBLESHOOTING.md#use-remote-root-directory-in-a-no-c-drive

kgermanov:
Ok, **Prefix Start Agent** resolved this

PR #434, opened 2023-09-07
Add Change Drive to CD command

JerekKimble:
If the jenkins agent drive location is on a different drive, then the CD doesnt change the drive and the following command fails to find the remoting.jar

### Submitter checklist


JerekKimble:
ah, breaks if the drive is the same now...

PR #435, opened 2023-09-07
Change Directory fails if the drive letter is different

JerekKimble:
The CD doesnt work if the drive doesn't match, resulting in the remoting.jar to not be found. the /D switch will cause the drive to change if it is different and do nothing if the drive matches.

### Submitter checklist


kuisathaverat:
the change of drive is documented https://github.com/jenkinsci/ssh-slaves-plugin/blob/main/doc/TROUBLESHOOTING.md#use-remote-root-directory-in-a-no-c-drive

kuisathaverat:
closed due to inactivity

PR #487, opened 2024-05-29
Updates Java 11 to Java 17 for test environment

alecharp:
### Testing done
When running with Java 17 and a Jenkins Core built with `-Dmaven.compiler.target=17` tests are failing. This resolves the problem.
As Jenkins will be compiled with such option in just 20days, this is providing from prevention.
As you can see from this pull request build result, there is no problem with the current core requirement and those changes.

### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/test/resources/hudson/plugins/sshslaves/agents/ssh-agent-curve25519-sha256/Dockerfile
+++ b/src/test/resources/hudson/plugins/sshslaves/agents/ssh-agent-curve25519-sha256/Dockerfile
@@ -14,7 +14,7 @@ RUN DEBIAN_FRONTEND="noninteractive" apt-get update -y -qq \
 RUN add-apt-repository ppa:openjdk-r/ppa -y \
   && apt-get update -y -qq \
   && apt-get install -y -qq \
-    openjdk-11-jdk \
+    openjdk-17-jdk \
   && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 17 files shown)


kuisathaverat (approved):

alecharp:
Thank you @kuisathaverat. As before, this will need to be released so the different testing environment can use the version.

kuisathaverat:
@alecharp Do test changes need a release? I do not think so; the tests are not included in the binary.

alecharp:
True, still they are used during PCT which will be broken until this patch is released.
As there is no changes in production files, releasing this patch is "risk-free" for the users.

kuisathaverat:
https://github.com/jenkinsci/ssh-agents-plugin/releases/tag/2.968.v6f8823c91de4

basil:
Indeed, https://github.com/jenkinsci/plugin-compat-tester works by cloning the tests for the repository at a particular release tag.

PR #542, opened 2024-12-20
fix: remove old code for PCT

kuisathaverat:
remove old code for PCT


kuisathaverat:
@basil How can I confirm that the PCT no longer needs this code?

basil:
I think only the code using the old terminology is unneeded. Testing can be done as described in https://github.com/jenkinsci/bom/blob/master/CONTRIBUTING.md#pct

PR #570, opened 2025-04-06
test: implement launcher using Apache Mina sshd library

kuisathaverat:
### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue


kuisathaverat:
Not sure about the CI errors

```
06:45:18  Caused by: org.apache.maven.surefire.booter.SurefireBooterForkException: There was an error in the forked process
06:45:18  [SUREFIRE] std/in stream corrupted
06:45:18  java.io.IOException: Stream closed
06:45:18  	at java.base/java.io.BufferedInputStream.getBufIfOpen(BufferedInputStream.java:168)
06:45:18  	at java.base/java.io.BufferedInputStream.read(BufferedInputStream.java:334)
06:45:18  	at java.base/java.io.BufferedInputStream.fill(BufferedInputStream.java:244)
06:45:18  	at java.base/java.io.BufferedInputStream.read1(BufferedInputStream.java:284)
06:45:18  	at java.base/java.io.BufferedInputStream.read(BufferedInputStream.java:343)
06:45:18  	at org.apache.maven.surefire.api.util.internal.Channels$3.readImpl(Channels.java:169)
06:45:18  	at org.apache.maven.surefire.api.util.internal.AbstractNoninterruptibleReadableChannel.read(AbstractNoninterruptibleReadableChannel.java:50)
06:45:18  	at org.apache.maven.surefire.api.stream.AbstractStreamDecoder.read(AbstractStreamDecoder.java:430)
06:45:18  	at org.apache.maven.surefire.api.stream.AbstractStreamDecoder.read(AbstractStreamDecoder.java:419)
06:45:18  	at org.apache.maven.surefire.api.stream.AbstractStreamDecoder.readMessageType(AbstractStreamDecoder.java:116)
06:45:18  	at org.apache.maven.surefire.booter.stream.CommandDecoder.decode(CommandDecoder.java:77)
06:45:18  	at org.apache.maven.surefire.booter.spi.CommandChannelDecoder.decode(CommandChannelDecoder.java:60)
06:45:18  	at org.apache.maven.surefire.booter.CommandReader$CommandRunnable.run(CommandReader.java:290)
06:45:18  	at java.base/java.lang.Thread.run(Thread.java:840)
06:45:18  
06:45:18      at org.apache.maven.plugin.surefire.booterclient.ForkStarter.fork (ForkStarter.java:628)
06:45:18      at org.apache.maven.plugin.surefire.booterclient.ForkStarter.lambda$null$7 (ForkStarter.java:401)
06:45:18      at java.util.concurrent.FutureTask.run (FutureTask.java:264)
06:45:18      at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1136)
06:45:18      at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:635)
06:45:18      at java.lang.Thread.run (Thread.java:840)
```

jglick:
Are any tests attempting to write to stdout/stderr (not using `System.out` / `System.err`)? I recall that for example https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/ProcessBuilder.html#inheritIO() can crash Surefire.

kuisathaverat:
I do not remember any, but I will review it.

kuisathaverat:
Jesse was right; the cause was a bunch of outputs to `System.err` and `System.out`. Now I have to resolbe a few SpotBugs warnings.

jglick:
Did you mean to update the PR title?

.editorconfig
```diff
@@ -9,6 +9,14 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
+[*.java]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
```
jglick, line 18: Recommend just enabling Spotless instead.

Jenkinsfile
```diff
@@ -6,5 +6,5 @@ buildPlugin(
   useContainerAgent: false, // TestContainers
   configurations: [
     [platform: 'linux', jdk: 21],
-    [platform: 'windows', jdk: 17],
+//    [platform: 'windows', jdk: 17],
```
jglick, line 9: Can this be reverted?

pom.xml
```diff
@@ -140,22 +160,97 @@
...
+    <pluginRepository>
+      <id>repo.jenkins-ci.org</id>
+      <url>https://repo.jenkins-ci.org/public/</url>
+    </pluginRepository>
+  </pluginRepositories>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>com.diffplug.spotless</groupId>
+        <artifactId>spotless-maven-plugin</artifactId>
```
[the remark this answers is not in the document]
kuisathaverat: https://github.com/jenkinsci/ssh-agents-plugin/pull/595
https://github.com/jenkinsci/ssh-agents-plugin/pull/596

Jenkinsfile
kuisathaverat: done

.editorconfig
kuisathaverat: Spotless is now enabled. I will keep these basic formatting rules here for all the IDEs

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/SSHApacheMinaLauncher.java
```diff
@@ -0,0 +1,1125 @@
...
+        if (validateHost.kind == FormValidation.Kind.ERROR) {
+            isValid = false;
+            message += validateHost.getMessage() + "\n";
+        }
+        if (validateCredentials.kind == FormValidation.Kind.ERROR) {
+            isValid = false;
+            message += validateCredentials.getMessage() + "\n";
+        }
+
+        if (!isValid) {
```
car-roll, line 607: Just a nit, but if the configuration is valid, won't the output just be:
`Validate configuration:`
with nothing else?

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/SSHApacheMinaLauncher.java
```diff
@@ -0,0 +1,1125 @@
...
+    /** Closes the SSH connection and any associated channels. */
+    private void close() {
+        try {
+            if (shellChannel != null) {
+                shellChannel.close();
+            }
+            if (connection != null) {
+                connection.close();
+            }
+        } catch (Exception e) {
```
car-roll, line 646: If `shellChannel.close()` throws an exception, should we still be trying to close `connection`?

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/SSHApacheMinaLauncher.java
```diff
@@ -0,0 +1,1125 @@
...
+    /**
+     * Enable/Disable the credential tracking, this tracking store information about where it is used
+     * a credential, in this case in a node. If the tracking is enabled and you launch a big number of
+     * Agents per day, activate credentials tacking could cause a performance issue see
+     *
+     * @see <a href= "https://issues.jenkins-ci.org/browse/JENKINS-49235">JENKINS-49235</a>
+     */
+    public boolean getTrackCredentials() {
+        String trackCredentials = System.getProperty(SSHApacheMinaLauncher.class.getName() + ".trackCredentials");
+        return !"false".equalsIgnoreCase(trackCredentials);
```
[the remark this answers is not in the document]
kuisathaverat: No, if the user is not able to put "true" or "false" in a system property, it is not setting the property correctly, so the property is invalid. You cannot manage all the wrong property values that a user can give you, and only make sense to fix their errors when they are common (which denotes there is another issue). 
To provide a `io.jenkins.plugins.sshbuildagents.ssh.mina.SSHApacheMinaLauncher.trackCredentials` with a value that you must trim. You have to provide something like `-Dio. Jenkins.plugins.sshbuildagents.ssh.mina.SSHApacheMinaLauncher.trackCredentials=" true "` that it is enough bizarre to do not thing anyone is so dumb.

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/Connection.java
```diff
@@ -0,0 +1,164 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016, Michael Clarke
```
olamy, line 4: Not sure why you use this copyright here? 
It looks to be a new class written by you. Either you put yourself or nobody but not someone who was not part of writing this code.

pom.xml
```diff
@@ -93,19 +93,48 @@
...
+          </exclusion>
+        </exclusions>
       </dependency>
     </dependencies>
   </dependencyManagement>
 
   <dependencies>
+    <dependency>
+      <groupId>io.jenkins.plugins</groupId>
+      <artifactId>eddsa-api</artifactId>
```
olamy, line 109: Could we remove this dependency and use this new feature from mina-sshd see https://github.com/apache/mina-sshd/pull/639
The project https://github.com/str4d/ed25519-java is not maintained anymore and have somen know security issues such https://github.com/str4d/ed25519-java/issues/95

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/SSHApacheMinaLauncher.java
```diff
@@ -0,0 +1,1125 @@
...
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void launch(@NonNull final SlaveComputer computer, @NonNull final TaskListener listener)
+            throws InterruptedException {
+        final Node node = computer.getNode();
+        final String host = this.host;
+        final int port = this.port;
+        if (computer == null || listener == null) {
```
olamy, line 322: not sure it's needed as both have `@NonNull ` annotation :)

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/Connection.java
kuisathaverat: C&P from the old class, so I keep the same header

kuisathaverat: I will review it, as it is likely new content, even though I started by copying it from the old file.

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/SSHApacheMinaLauncher.java
kuisathaverat: I struggle to remove a warning from SpotBugs; this is a leftover. I will remove it.

kuisathaverat:
> Nice PR. Some questions:
> 
> * should we maintain both implementations?

At least until we are confident with this new launcher, it lacks some features the other has, like host verification. 

>   Maybe it could be more simple (and less maintenance) to have a single implementation based on Apache Mina?
> * Maybe we could get rid of trilead as well?

That's the goal; I will stop accepting and processing changes in Trilead because I will focus my time on this new launcher and a way to migrate the old configuration to the latest in a seamless manner. It will take a while for sure.

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/ConnectionImpl.java
```diff
@@ -0,0 +1,411 @@
...
+     * Connets to the SSH service configured and authenticate
+     *
+     * @return Returns a ClientSession connected and authenticated.
+     * @throws IOException in case of error.
+     */
+    private ClientSession connectAndAuthenticate() throws IOException {
+        ConnectFuture connectionFuture = client.connect(this.credentials.getUsername(), this.host, this.port);
+        connectionFuture.verify(this.timeoutMillis);
+        session = connectionFuture.getSession();
+        addAuthentication();
```
olamy, line 267: you should be able to use something 
```
SSHAuthenticator.newInstance(session, credentials).authenticate(TaskListener.NULL)
```

olamy (dismissed):

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/ServerHostKeyVerifier.java
```diff
@@ -0,0 +1,32 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016, Michael Clarke
```
olamy, line 4: ditto

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/ShellChannel.java
```diff
@@ -0,0 +1,70 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016, Michael Clarke
```
olamy, line 4: ditto

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/FakeURI.java
```diff
@@ -0,0 +1,26 @@
+package io.jenkins.plugins.sshbuildagents.ssh.mina;
```
olamy, line 1: missing license but we probably will not need this file anymore

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/KeyAlgorithmManagerImpl.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016, Michael Clarke
```
olamy, line 4: ditto

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/PrivateKeyParser.java
```diff
@@ -0,0 +1,89 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016, Michael Clarke
```
olamy, line 4: ditto.

but not sure we really need this file

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/SSHApacheMinaLauncher.java
```diff
@@ -0,0 +1,1125 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2004-, all the contributors
```
olamy, line 4: yeah :)

src/test/java/io/jenkins/plugins/sshbuildagents/ssh/ConnectionImplTest.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    @Test
+    public void testRunLongConnection() throws IOException, InterruptedException {
+        Connection connection = new ConnectionImpl(sshd.getHost(), sshd.getPort());
+        StandardUsernameCredentials credentials = new FakeSSHKeyCredential();
+        connection.setCredentials(credentials);
+        ShellChannel shellChannel = connection.shellChannel();
+        shellChannel.execCommand("sleep 500s");
+        for (int i = 0; i < 300; i++) {
```
olamy, line 133: thanks that's plenty of time to have a coffee :)
do we really need this? Can't we skip this on normal dev life and run it on CI server only?

olamy:
right I have started to put my comments into code with this https://github.com/kuisathaverat/ssh-agents-plugin/pull/102 
obviously some tests are failing as now ConnectionImpl is using the existing authz mode which doesn't work without jenkins

olamy:
> I will address some issues and then merge them as is. This PR is a WIP new launcher, and this is the first iteration. It is not perfect, and it should not be. I do not want to make this PR an open issue that remains unmerged. Also, it helps to work on different improvements in parallel and track the changes better

If you do so, you must at least mark and inform users about the experimental aspect of your PR when they use the new launcher because many features are not working now.

I completely understand the rationale behind adding an abstraction layer on top of the SSH connection. In theory, it makes sense and can offer flexibility.
That said, I'm wondering if it's really necessary in the specific context of the Jenkins SSH Agents plugin. Do we foresee switching from Apache Mina to another implementation anytime soon (while it took already so long having some changes out of trilead :))? From what I can see, both the SSH protocol and the Apache Mina library are quite stable and well-maintained, especially when it comes to bugs and security.

Introducing a new abstraction would mean committing to a public API that we'll need to support in the long term. Given that this is currently scoped just to the Jenkins SSH Agents plugin. I'm not sure the added complexity is justified.

Also, I'm a bit hesitant about re-implementing the authentication layer from scratch. The existing solution is widely used, battle-tested, and already aligns with security requirements. 
 https://github.com/jenkinsci/mina-sshd-api-plugin/tree/main/mina-sshd-api-core/src/main/java/io/jenkins/plugins/mina_sshd_api/core/authenticators



image

I wish to have you open to comments as definitely such changes need some discussion and acceptance by the broader community, as this has some large effect on the long term maintenance of the plugin

olamy (dismissed):

src/test/java/io/jenkins/plugins/sshbuildagents/ssh/agents/AgentConnectionBaseTest.java
```diff
@@ -0,0 +1,158 @@
...
+
+    protected JenkinsRule j;
+
+    @BeforeEach
+    void beforeEach(JenkinsRule j) {
+        this.j = j;
+        this.j.timeout = 0;
+    }
+
+    @BeforeAll
```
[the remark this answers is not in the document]
kuisathaverat: yes, that should be enough

src/test/java/io/jenkins/plugins/sshbuildagents/ssh/agents/AgentConnectionBaseTest.java
```diff
@@ -0,0 +1,158 @@
...
+    static void beforeAll() {
+        assumeFalse(SystemUtils.IS_OS_WINDOWS);
```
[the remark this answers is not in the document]
kuisathaverat: it is another way to do it

kuisathaverat:
> > I will address some issues and then merge them as is. This PR is a WIP new launcher, and this is the first iteration. It is not perfect, and it should not be. I do not want to make this PR an open issue that remains unmerged. Also, it helps to work on different improvements in parallel and track the changes better
> 
> If you do so, you'll need to mark and let users know about the experimental aspect of your PR when they use the new launcher because many features are not working now (maybe some warning as well when using this using the new launcher)
> 

Yes, it will be marked as experimental, not production-ready in the UI before I merge.

> I completely understand the rationale behind adding an abstraction layer on top of the SSH connection. In theory, it makes sense and can offer flexibility. That said, I'm wondering if it's really necessary in the specific context of the Jenkins SSH Agents plugin. Do we foresee switching from Apache Mina to another implementation anytime soon (while it took already so long having some changes out of trilead :))? From what I can see, both the SSH protocol and the Apache Mina library are quite stable and well-maintained, especially when it comes to bugs and security.
> 
> Introducing a new abstraction would mean committing to a public API that we'll need to support in the long term. Given that this is currently scoped just to the Jenkins SSH Agents plugin. I'm not sure the added complexity is justified.
> 

I have a different opinion that the abstraction layer makes it easy to adapt the Apache Mina SSH library to the terms of Jenkins, making it easy to troubleshoot and locate errors in the implementation. Makes the Launcher more modular, making it easy to understand what each part does. Each part functions as a black box, performing a single task. To have an abstraction layer does not affect security or updates from the Apache Mina SSH library. We will continue to update the library with new releases.

> Also, I'm a bit hesitant about re-implementing the authentication layer from scratch. The existing solution is widely used, battle-tested, and already aligns with security requirements. https://github.com/jenkinsci/mina-sshd-api-plugin/tree/main/mina-sshd-api-core/src/main/java/io/jenkins/plugins/mina_sshd_api/core/authenticators
> 

At the time I started this launcher, that code did not exist, I would evaluate if I can reuse it, but not in this PR.

> I wish to have you open to comments as definitely such changes need some discussion and acceptance by the broader community, as this has some large effect on the long term maintenance of the plugin

I am open to any suggestions you have. I have been part of the Jenkins community for over 10 years, maintaining six active plugins and having previously maintained four more. My plugins are all well-maintained and up-to-date, and I am always open to accepting contributions and suggestions. The Jenkins community is nothing new to me. I have always been an active member of the community.

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/Connection.java
```diff
@@ -0,0 +1,164 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+import org.apache.sshd.client.session.ClientSession;
+
+/**
+ * Interface to manage an SSH connection.
+ *
+ * @author Ivan Fernandez Calvo
+ */
+public interface Connection extends AutoCloseable {
```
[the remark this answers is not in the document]
kuisathaverat: Currently, I see more benefits to using an abstraction layer on top of Apache Mina SSH.

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/ConnectionImpl.java
kuisathaverat: I am testing it

olamy:
> > > I will address some issues and then merge them as is. This PR is a WIP new launcher, and this is the first iteration. It is not perfect, and it should not be. I do not want to make this PR an open issue that remains unmerged. Also, it helps to work on different improvements in parallel and track the changes better
> > 
> > 
> > If you do so, you'll need to mark and let users know about the experimental aspect of your PR when they use the new launcher because many features are not working now (maybe some warning as well when using this using the new launcher)
> 
> Yes, it will be marked as experimental, not production-ready in the UI before I merge.
> 

Maybe changing the PR title as well?

> > I completely understand the rationale behind adding an abstraction layer on top of the SSH connection. In theory, it makes sense and can offer flexibility. That said, I'm wondering if it's really necessary in the specific context of the Jenkins SSH Agents plugin. Do we foresee switching from Apache Mina to another implementation anytime soon (while it took already so long having some changes out of trilead :))? From what I can see, both the SSH protocol and the Apache Mina library are quite stable and well-maintained, especially when it comes to bugs and security.
> > Introducing a new abstraction would mean committing to a public API that we'll need to support in the long term. Given that this is currently scoped just to the Jenkins SSH Agents plugin. I'm not sure the added complexity is justified.
> 
> I have a different opinion that the abstraction layer makes it easy to adapt the Apache Mina SSH library to the terms of Jenkins, making it easy to troubleshoot and locate errors in the implementation. Makes the Launcher more modular, making it easy to understand what each part does. Each part functions as a black box, performing a single task. To have an abstraction layer does not affect security or updates from the Apache Mina SSH library. We will continue to update the library with new releases.
> 

Not sure you got my point. I mean I'm not sure there is need for an interface Connection and ConnectionImpl and especially make this a public API because at the end of day there will be only a single implementation.

Why not simply make a single class MinaConnection package private so it will not be seen as a public API to maintain. same for other new interfaces.


> > Also, I'm a bit hesitant about re-implementing the authentication layer from scratch. The existing solution is widely used, battle-tested, and already aligns with security requirements. https://github.com/jenkinsci/mina-sshd-api-plugin/tree/main/mina-sshd-api-core/src/main/java/io/jenkins/plugins/mina_sshd_api/core/authenticators
> 
> At the time I started this launcher, that code did not exist, I would evaluate if I can reuse it, but not in this PR.
> 
> > I wish to have you open to comments as definitely such changes need some discussion and acceptance by the broader community, as this has some large effect on the long term maintenance of the plugin
> 
> I am open to any suggestions you have. I have been part of the Jenkins community for over 10 years, maintaining six active plugins and having previously maintained four more. My plugins are all well-maintained and up-to-date, and I am always open to accepting contributions and suggestions. The Jenkins community is nothing new to me. I have always been an active member of the community.

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/ConnectionImpl.java
```diff
@@ -0,0 +1,363 @@
...
+ * Implements {@link Connection} using the Apache Mina SSHD library
+ * https://github.com/apache/mina-sshd
+ *
+ */
+public class ConnectionImpl implements Connection {
+    /** The number of heartbeat packets lost before closing the connection. */
+    public static final int HEARTBEAT_MAX_RETRY = 6;
+
+    /** The size of the SSH window size in bytes. */
+    public static final long WINDOW_SIZE = 4L * 1024 * 1024;
```
olamy, line 39: how did you come to those values? As they are very different from the default Mina values, maybe this would be nice to offer the possibility to override via `SystemProperties`

olamy (approved):

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/ConnectionImpl.java
```diff
@@ -0,0 +1,363 @@
...
+        session = connectionFuture.getSession();
+        var authenticator = SSHAuthenticator.newInstance(session, credentials);
+        authenticator.authenticate(TaskListener.NULL);
+        return session;
+    }
+
+    /** Initialize the SSH client. It reuses the client if it exists. */
+    private void initClient() {
+        if (client == null) {
+            client = SshClient.setUpDefaultClient();
```
olamy, line 226: SshClient is meant to be reuse to avoid excessive allocation. 
We can have a single instance, I have implemented this https://github.com/olamy/ssh-agents-plugin/commit/d87a31cc3afd71a5b0d67819c9df84e1371d68c0
saddly you have forced push to your branch which has totally broken the branch and PR to your repo

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/ConnectionImpl.java
```diff
@@ -0,0 +1,363 @@
...
+            CoreModuleProperties.WINDOW_SIZE.set(client, WINDOW_SIZE);
+            CoreModuleProperties.TCP_NODELAY.set(client, tcpNoDelay);
+            CoreModuleProperties.HEARTBEAT_REQUEST.set(client, "keepalive@jenkins.io");
```
olamy, line 229: using KeepAliveHandler would have the advantage about testing the application (ssh) level and not only low level 
see commit here https://github.com/olamy/ssh-agents-plugin/commit/f3f3b6d3d32ce46226cb1aa169c73737e6b92390 
the KeepAlive have some benefits such:
- SSH protocol-level compliant so the server is expected to respond.
- will help detect unreachable servers 
- MINA can close the session automatically if the server doesn’t respond.

olamy (changes requested):

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/KeyAlgorithmManagerImpl.java
```diff
@@ -0,0 +1,22 @@
...
+package io.jenkins.plugins.sshbuildagents.ssh.mina;
+
+import io.jenkins.plugins.sshbuildagents.ssh.KeyAlgorithm;
+import io.jenkins.plugins.sshbuildagents.ssh.KeyAlgorithmManager;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Implementation of KeyAlgorithmManager that provides a list of supported key algorithms. TODO
+ * Implement a proper key algorithm manager that returns supported algorithms.
```
olamy, line 14: what will be the goal of this? 
Maybe it could be just as easy as delegates this Mina as anyway, we will depend on his validation mechanism.

olamy (changes requested):

pom.xml
```diff
@@ -104,11 +104,6 @@
   </dependencyManagement>
 
   <dependencies>
-    <!-- TODO remove eddsa-api library ans us the new feature of Apache Mina SSH https://github.com/apache/mina-sshd/pull/639 -->
```
olamy: Unfortunately, this won't be able to be done yet. 
At least until the plugin still supports the trilead library, or this will break users who haven't migrated and needs eddsa support.

olamy (changes requested):

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/ConnectionImpl.java
kuisathaverat: You can configure the keepalive option in the SSH server. I need to verify whether enabling it in the client affects the configuration you set on the server.

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/ConnectionImpl.java
olamy: Right so make it configurable can be a new feature in a follow up PR.

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/ConnectionImpl.java
olamy: Why? Can you explain your reasoning? 
This is very different from what is recommended by the Apache Mina project itself. 
See javadoc https://javadoc.io/doc/org.apache.sshd/sshd-core/2.14.0/org/apache/sshd/client/SshClient.html
For installation with large number of agents, this will consume/create a lot resources (such threadpool, io etc..) while it's not needed.

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/ConnectionImpl.java
felfert: > especially this
> 
> ```
> This configuration only ensures that the server does not terminate the session due to no traffic. If the incoming traffic from the server may also suffer from long "quiet" periods, one runs the risk of a client time-out.
> ```

Which is not entirely true. It also helps to **detect** a broken (for whatever reason) connection, and **THAT** is the important part. Regardless, a server can terminate the the connection at any time on the application level.

olamy: > > especially this
> > ```
> > This configuration only ensures that the server does not terminate the session due to no traffic. If the incoming traffic from the server may also suffer from long "quiet" periods, one runs the risk of a client time-out.
> > ```
> 
> Which is not entirely true. It also helps to **detect** a broken (for whatever reason) connection, and **THAT** is the important part. Regardless, a server can terminate the the connection at any time on the application level.

not sure to understand? Do you mean documentation from Apache Mina is wrong?
My point is we can have both as it doesn't hurt much (or having it optional but activable via SystemProperties).

src/test/java/io/jenkins/plugins/sshbuildagents/ssh/agents/AgentConnectionBaseTest.java
```diff
@@ -0,0 +1,166 @@
...
+        boolean ret = false;
+        int count = 0;
+        while (count < 30) {
+            Thread.sleep(1000);
+            String log = node.toComputer().getLog();
+            ret = log.contains("Agent successfully connected and online");
+            count++;
+        }
+        return ret;
+    }
```
jtnord, line 99: did you mean to just waste time spinning here up to 30 seconds if the agent connects after 1 second?

```suggestion
    protected static boolean isSuccessfullyConnected(Node node) throws IOException, InterruptedException {
        int count = 0;
        while (count < 30) {
            Thread.sleep(1000);
            String log = node.toComputer().getLog();
            if (log.contains("Agent successfully connected and online")) {
              return true;
            }
            count++;
        }
        return false;
    }
```

src/test/java/io/jenkins/plugins/sshbuildagents/ssh/agents/AgentRSA512ConnectionTest.java
```diff
@@ -0,0 +1,45 @@
...
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.junit.jupiter.Container;
+
+/**
+ * Connect to a remote SSH Agent
+ *
+ */
+@Tag(AGENT_SSH_TEST)
+@Tag(SSH_KEX_TEST)
+public class AgentRSA512ConnectionTest extends AgentConnectionBaseTest {
```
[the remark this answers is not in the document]
jtnord: found the issue with the test taking too long, see comment in src/test/java/io/jenkins/plugins/sshbuildagents/ssh/agents/AgentConnectionBaseTest.java

src/test/java/io/jenkins/plugins/sshbuildagents/ssh/agents/ClientRSA512ConnectionTest.java
```diff
@@ -0,0 +1,176 @@
...
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+/**
+ * Connect to a remote SSH Server with a plain Apache Mina SSHD client.
+ *
+ */
+@Tag(AGENT_SSH_TEST)
+@Tag(SSH_KEX_TEST)
+@Testcontainers(disabledWithoutDocker = true)
+@DisabledOnOs(OS.WINDOWS)
```
jtnord, line 55: presumably because of the containers on windows in CI and windows containers and not Linux?

generally better to use a more specific exclusion (here and elsewhere).
e.g. (junit4) https://github.com/jenkinsci/ldap-plugin/pull/47

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/SSHApacheMinaLauncher.java
```diff
@@ -0,0 +1,1107 @@
...
+                                ? sshHostKeyVerificationStrategy.getClass().getName()
+                                : "None");
+        sb.append(", tcpNoDelay=").append(getTcpNoDelay());
+        sb.append(", trackCredentials=").append(getTrackCredentials());
+        sb.append('}');
+        return sb.toString();
+    }
+
+    @Extension
+    @Symbol({"ssh", "sshMinaLauncher"})
```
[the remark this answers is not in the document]
jglick: rather
```suggestion
    @Symbol("sshMina")
```
(see Javadoc for `@Symbol`)

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/Connection.java
```diff
@@ -0,0 +1,144 @@
...
+     */
+    String getHostname();
+
+    /**
+     * @return Return the port configured to connect by SSH.
+     */
+    int getPort();
+
+    /** Close the connections and resources associated. */
+    void close();
```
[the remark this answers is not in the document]
kuisathaverat: I probably can remove it, it is already defined in the interface

src/test/java/io/jenkins/plugins/sshbuildagents/ssh/agents/AgentConnectionBaseTest.java
kuisathaverat: True, I did not check ret to exit from the loop

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/Connection.java
```diff
@@ -0,0 +1,144 @@
...
+     * @return The error code returned by the command.
+     * @throws IOException in case of error.
+     */
+    int execCommand(String command) throws IOException;
+
+    /**
+     * Create a {@link #shellChannel()} to execute non-interactive commands.
+     *
+     * @return Return a {@link #shellChannel()}
+     * @throws IOException
```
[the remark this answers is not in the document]
kuisathaverat: Having Javadoc does not hurt anyone; not having it keeps things undocumented.

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/ConnectionImpl.java
```diff
@@ -0,0 +1,363 @@
...
+    public static final int IDLE_SESSION_TIMEOUT = 60;
+
+    /** The standard output stream of the channel. */
+    private OutputStream stdout = System.out;
+
+    /** The standard error stream of the channel. */
+    private OutputStream stderr = System.err;
+
+    /** The SSH client used for the connection. */
+    private SshClient client;
```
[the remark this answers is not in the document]
kuisathaverat: This launcher is a WIP. In the first implementation, it is as it is. In the next iteration, I will implement and test the pool, but not in the first iteration.

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/ConnectionImpl.java
```diff
@@ -0,0 +1,363 @@
...
+     * @param port The port number of the SSH server.
+     */
+    public ConnectionImpl(String host, int port) {
+        this.host = host;
+        this.port = port;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public int execCommand(String command) throws IOException {
```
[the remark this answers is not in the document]
kuisathaverat:
```suggestion
// FIXME review, it does not return the exit code as it is said in the javadoc
    public int execCommand(String command) throws IOException {
```

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/ConnectionImpl.java
```diff
@@ -0,0 +1,362 @@
...
+        client = null;
+        session = null;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void copyFile(String remotePath, byte[] bytes, boolean overwrite, boolean checkSameContent)
+            throws IOException {
+        try (ClientSession session = connect()) {
+            DefaultScpClient scp = new DefaultScpClient(session);
```
kuisathaverat, line 149:
```suggestion
// TODO set access/modification time to be the same as the source
            DefaultScpClient scp = new DefaultScpClient(session);
```

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/ConnectionImpl.java
```diff
@@ -0,0 +1,362 @@
...
+    public ClientSession connect() throws IOException {
+        initClient();
+        if (isSession() == false) {
+            for (int i = 0; i <= maxNumRetries; i++) {
+                try {
+                    return connectAndAuthenticate();
+                } catch (Exception ex) {
+                    String message = getExMessage(ex);
+                    if (maxNumRetries - i > 0) {
+                        println(
```
kuisathaverat, line 180:
```suggestion
// FIXME send these logs to the TaskListener
                        println(
```

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/ConnectionImpl.java
```diff
@@ -0,0 +1,362 @@
...
+     * @return Returns a ClientSession connected and authenticated.
+     * @throws IOException in case of error.
+     * @throws InterruptedException
+     */
+    private ClientSession connectAndAuthenticate() throws IOException, InterruptedException {
+        ConnectFuture connectionFuture = client.connect(this.credentials.getUsername(), this.host, this.port);
+        connectionFuture.verify(this.timeoutMillis);
+        session = connectionFuture.getSession();
+        var authenticator = SSHAuthenticator.newInstance(session, credentials);
+        authenticator.authenticate(TaskListener.NULL);
```
kuisathaverat, line 218:
```suggestion
// FIXME send these logs to the TaskListener of the computer launcher
        authenticator.authenticate(TaskListener.NULL);
```

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/ConnectionImpl.java
```diff
@@ -0,0 +1,362 @@
...
+
+    /** {@inheritDoc} */
+    @Override
+    public void setWorkingDirectory(String path) {
+        this.workingDirectory = path;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void setStdErr(OutputStream stderr) {
```
kuisathaverat, line 329:
```suggestion
// FIXME rename for Stderr
    public void setStdErr(OutputStream stderr) {
```

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/ShellChannelImpl.java
```diff
@@ -0,0 +1,114 @@
...
+import org.apache.sshd.client.channel.ChannelSession;
+import org.apache.sshd.client.channel.ClientChannelEvent;
+import org.apache.sshd.client.session.ClientSession;
+
+/**
+ * Implements {@link ShellChannel} using the Apache Mina SSHD library
+ * https://github.com/apache/mina-sshd
+ *
+ */
+public class ShellChannelImpl implements ShellChannel {
```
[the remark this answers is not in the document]
kuisathaverat:
```suggestion
// TODO think to rename to ShellRunner or something like that, there are classes in Mina with a similar name
public class ShellChannelImpl implements ShellChannel {
```

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/Connection.java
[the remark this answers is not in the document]
jtnord: > Having Javadoc does not hurt anyone; not having it keeps things undocumented.

But my point is this as is this adds no documentation.
In fact it is worse as you don't even get a warning saying you haven't documented it so you can not keep tracked of things missing documentation correctly

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/Connection.java
```diff
@@ -0,0 +1,144 @@
...
+    int getPort();
+
+    /** Close the connections and resources associated. */
+    void close();
+
+    /**
+     * Copy a file to the host by SCP. It does not create folders, so the folders of the path should
+     * exist.
+     *
+     * @param remoteFile Full path to the remote file.
```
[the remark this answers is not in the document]
kuisathaverat: yep, probably it can

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/SSHApacheMinaLauncher.java
```diff
@@ -0,0 +1,1107 @@
...
+
+    /**
+     * Reports the environment variables for the remote user on the Agent logs.
+     *
+     * @param listener The listener to report the environment variables to.
+     * @throws IOException If an error occurs while reporting the environment variables.
+     */
+    protected void reportEnvironment(TaskListener listener) throws IOException {
+        listener.getLogger().println(Messages._SSHLauncher_RemoteUserEnvironment(getTimestamp()));
+        connection.execCommand("set");
```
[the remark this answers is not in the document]
kuisathaverat: `set` works on Unix and Windows

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/SSHApacheMinaLauncher.java
```diff
@@ -0,0 +1,1107 @@
...
+        connection.setStdOut(new NoCloseOutputStream(listener.getLogger()));
+        connection.connect();
+    }
+
+    /**
+     * Validates the Agent configuration.
+     *
+     * @throws InterruptedException
+     */
+    private void checkConfig() throws InterruptedException {
```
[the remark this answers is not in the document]
kuisathaverat: sure, there is a bunch of checks in the form, but this ensures the runtime configuration is valid. This method resolved a lot of issues in the old version of the launcher. It is a check check that resolves user issues.

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/SSHApacheMinaLauncher.java
```diff
@@ -0,0 +1,1107 @@
...
+
+        /**
+         * Checks if the given java path is valid.
+         *
+         * @param value The java path to check.
+         * @return A FormValidation indicating whether the java path is valid or not.
+         */
+        @RequirePOST
+        public FormValidation doCheckJavaPath(@QueryParameter String value) {
+            FormValidation ret = FormValidation.ok();
```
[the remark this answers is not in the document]
kuisathaverat: Copy & paste from the old launcher there is room to improbe it, this is the initial implementation

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/SSHApacheMinaLauncher.java
```diff
@@ -0,0 +1,1107 @@
...
+     * @param agent The agent to get the root workspace of.
+     * @return the remote root workspace (without trailing slash).
+     */
+    @CheckForNull
+    private static String getWorkingDirectory(@CheckForNull Slave agent) {
+        if (agent == null) {
+            return null;
+        }
+        String workingDirectory = agent.getRemoteFS();
+        while (workingDirectory.endsWith("/")) {
```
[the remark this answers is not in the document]
kuisathaverat: good for them

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/SSHApacheMinaLauncher.java
```diff
@@ -0,0 +1,1107 @@
...
+
+                return merged;
+            } else {
+                return global;
+            }
+        } else if (local != null) {
+            return local;
+        } else {
+            return new EnvVars();
+        }
```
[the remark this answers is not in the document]
kuisathaverat: Copy and Paste from the Old launcher

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/SSHApacheMinaLauncher.java
```diff
@@ -0,0 +1,1107 @@
...
+     * set.
+     *
+     * @return The SSH host key verification strategy.
+     */
+    @NonNull
+    SshHostKeyVerificationStrategy getSshHostKeyVerificationStrategyDefaulted() {
+        return sshHostKeyVerificationStrategy != null
+                ? sshHostKeyVerificationStrategy
+                : new NonVerifyingKeyVerificationStrategy();
+    }
```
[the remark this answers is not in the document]
kuisathaverat: no host verification yet

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/SSHApacheMinaLauncher.java
```diff
@@ -0,0 +1,1107 @@
...
+     *
+     * @return Value for property 'host'.
+     */
+    public String getHost() {
+        return host;
+    }
+
+    public void setHost(String value) {
+        this.host = Util.fixEmptyAndTrim(value);
+    }
```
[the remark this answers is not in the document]
kuisathaverat: the host is validated in the descriptor, this only trim the value that's all

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/SSHApacheMinaLauncher.java
```diff
@@ -0,0 +1,1107 @@
...
+                        // Collections.singletonList(SSH_SCHEME),
+                        Collections.singletonList(new HostnamePortRequirement(host, portValue)),
+                        SSHAuthenticator.matcher(ClientSession.class))) {
+                    if (StringUtils.equals(value, o.value)) {
+                        return FormValidation.ok();
+                    }
+                }
+            } catch (NumberFormatException e) {
+                return FormValidation.warning(e, Messages.SSHLauncher_PortNotANumber());
+            }
```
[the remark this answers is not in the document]
kuisathaverat: this code is more or less the same you can find at the Jenkins documentation abotu how to validate form values
https://www.jenkins.io/doc/developer/forms/form-validation/#getting-values-of-other-fields

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/SSHApacheMinaLauncher.java
```diff
@@ -0,0 +1,1107 @@
...
+        }
+
+        /**
+         * Checks if the given port is valid.
+         *
+         * @param value The port to check.
+         * @return A FormValidation indicating whether the port is valid or not.
+         */
+        @RequirePOST
+        public FormValidation doCheckPort(@QueryParameter String value) {
```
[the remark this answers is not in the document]
kuisathaverat: event though, we valudate in the Jelly form, it must validate in Java code that everthing is correct.

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/verifiers/KnowHostServerKeyVerifier.java
```diff
@@ -0,0 +1,19 @@
...
+
+package io.jenkins.plugins.sshbuildagents.ssh.mina.verifiers;
+
+import org.apache.sshd.client.keyverifier.ServerKeyVerifier;
+
+/**
+ * A server key verifier that uses the known hosts file for verification.
+ *
+ */
+public class KnowHostServerKeyVerifier extends KeyVerifier {
```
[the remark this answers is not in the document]
kuisathaverat: WIP, host verification is not implemented yet

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/verifiers/RequiredServerKeyVerifier.java
```diff
@@ -0,0 +1,36 @@
...
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.security.PublicKey;
+import org.apache.sshd.client.keyverifier.ServerKeyVerifier;
+
+/**
+ * A server key verifier that requires a specific public key for authentication.
+ *
+ */
+public class RequiredServerKeyVerifier extends KeyVerifier {
```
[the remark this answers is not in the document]
kuisathaverat: WIP

src/main/resources/io/jenkins/plugins/sshbuildagents/ssh/mina/SSHApacheMinaLauncher/config.jelly
```diff
@@ -0,0 +1,62 @@
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout"
+         xmlns:t="/lib/hudson" xmlns:f="/lib/form" xmlns:c="/lib/credentials">
+  <div class="jenkins-alert jenkins-alert-danger">
+    The SSH Apache Mina Launcher is in early state of development and is not recommended for production use.
+  </div>
+
+  <f:entry title="${%Host}" field="host">
+    <f:textbox checkMethod="post"/>
```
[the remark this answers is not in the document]
kuisathaverat: cool, you are more updated than me, you work for CloudBees.

src/main/resources/io/jenkins/plugins/sshbuildagents/Messages.properties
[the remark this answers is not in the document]
kuisathaverat: historic, not related with this PR.

src/main/java/io/jenkins/plugins/sshbuildagents/ssh/mina/ConnectionImpl.java
```diff
@@ -0,0 +1,363 @@
...
+    public static final long WINDOW_SIZE = 4L * 1024 * 1024;
+
+    /** The time in seconds to wait before sending a keepalive packet. */
+    public static final int HEARTBEAT_INTERVAL = 10;
+
+    /** The time in minutes to wait before closing the session if no command is executed. */
+    public static final int IDLE_SESSION_TIMEOUT = 60;
+
+    /** The standard output stream of the channel. */
+    private OutputStream stdout = System.out;
```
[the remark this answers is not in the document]
kuisathaverat: I will think about it

